PR #1, opened 2017-07-07
Added an identifier arg

mbrandonw:
Added an optional `identifier:` arg to the assertion so that you can differentiate between multiple assertions in the same function.

I also trimmed the `()` off the function for the screenshot name.


stephencelis:
Gonna merge and work on some stuff!

PR #7, opened 2017-07-09
Encodable support

stephencelis:
Allows us to snap against any `Encodable` type. Uses `JSONEncoder` under the hood right now, but we probably want to use our own encoder in the future.


Sources/SnapshotAssertion/SnapshotAssertion.swift
```diff
@@ -119,3 +120,25 @@ public func assertSnapshot<S: Snapshot>(
     return
   }
 }
+
+public func assertSnapshot<S: Encodable>(
+  encoding snapshot: S,
```
stephencelis: Changed the keyword here since there's ambiguity:

``` swift
assertSnapshot(matching: data)
// vs.
assertSnapshot(encoding: data)
```

Sources/SnapshotAssertion/SnapshotAssertion.swift
```diff
@@ -72,9 +72,10 @@ extension String: Snapshot {
 public func assertSnapshot<S: Snapshot>(
   matching snapshot: S,
   identifier: String? = nil,
-  _ file: StaticString = #file,
-  _ function: String = #function,
-  _ line: UInt = #line)
+  pathExtension: String? = nil,
```
mbrandonw: nice nice! so this will allow me to override extension when doing a plain text snapshot?

stephencelis: Yep!

PR #15, opened 2017-07-09
Build macOS on Circle

stephencelis:
Should help catch some things.


stephencelis:
Well I got the NSImage retina mismatch issues resolved both for rendering on non-retina (_e.g._ Circle) and comparing. Kinda nasty stuff, but we can attempt to clean up later :/

stephencelis:
Snuck in a fix for #13 on this branch (since there are a lot of changes here).

screen shot 2017-07-09 at 1 01 56 pm

Sources/SnapshotTesting/Cocoa.swift
```diff
@@ -12,33 +12,50 @@
     }
 
     public var diffableData: Data {
+      let scale = NSScreen.main!.backingScaleFactor
+      self.size = .init(width: self.size.width * 2.0 / scale, height: self.size.height * 2.0 / scale)
```
stephencelis: This ensures the size on non-retina displays matches that of retina displays.

Sources/SnapshotTesting/Cocoa.swift
```diff
@@ -12,33 +12,50 @@
...
       let rep = NSBitmapImageRep(cgImage: cgImage)
       rep.size = self.size
       let data = rep.representation(using: .png, properties: [:])!
       return data
     }
 
-    public func diff(comparing other: Data) -> XCTAttachment? {
-      let existing = NSImage(data: other)!
+    public func diff(from other: NSImage) -> Bool {
+      let bitmapInfo = NSImage(data: self.diffableData)!
```
stephencelis: Have to convert the image to data and back again for the `bitmapInfo`s to match :(

Up for suggestions on optimizing here.

PR #18, opened 2017-08-01
Basics of scenekit snapshots

mbrandonw:
Made `SCNScene` snapshot capable by creating an `SCNView`, sticking in the scene, and then SceneKit has first class support for taking a snapshot.

unfortunately you have to bake the size of the snapshot into the `snapshotFormat` function, which probably isn't write. I think it's prob way more likely to create your own scene view, slap in the scene, and then snapshot it. but whatev, this was fun anyway.


mbrandonw:
lol gonna close, will pick up once we get custom snapshotters

Issue #19, opened 2017-08-02
make subdirectories for each test class


mbrandonw:
will help for lots of tests

Issue #20, opened 2017-08-02
automatic number for multiple tests in a single test case


mbrandonw:
helpful for ordering files

PR #21, opened 2017-08-10
Cleanup, etc.

stephencelis:
Big, breaking changes.

1. Got rid of `assertSnapshot(encoding:)`—didn't seem worth it. The `Any` check is more powerful and strict.
2. Got rid of the trailing closure `record { … }` API and instead added `record: true` to the assertion itself.
3. Recording always causes a test assertion failure. This lets us quickly check what was recorded by looking at the Xcode error and avoids letting unrecorded assertions get checked in and pass CI.
4. String diff failures use a new diff format. This is easier to read than a general call to `XCTAssertEqual`.
    screen shot 2017-08-09 at 11 40 14 pm
5. Snapshots are now grouped at the test file level (rather than test directory level). This means more directory groupings with shorter file names that are easier to scan in GitHub pull requests.
6. Multiple snapshots in a single test will increment automatically. This makes sorting through snapshots chronologically easier. It all means setting a snapshot name is optional.


stephencelis:
(This fixes #19 and #20 when merged.)

PR #26, opened 2017-09-02
Show truncated diff context

stephencelis:
Uses a patch-like format to show truncated, focussed diffs on failure. _E.g._,

screen shot 2017-09-02 at 6 57 54 pm

### DIFF
--- a/Sources/Diff/Diff.swift
+++ b/Sources/Diff/Diff.swift
@@ -1,7 +1,12 @@
-public enum Diff<A> {
-  case fst(A)
-  case snd(A)
-  case tup(A, A)
+public enum DiffType {
+  case first
+  case second
+  case both
+}
+
+public struct Diff<A> {
+  public let type: DiffType
+  public let elements: [A]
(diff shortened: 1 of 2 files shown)


Sources/Diff/Diff.swift
```diff
@@ -24,10 +29,85 @@ public func diff<A: Hashable>(_ fst: [A], _ snd: [A]) -> [Diff<A>] {
...
+}
+
+public func chunk<S: StringProtocol>(diff diffs: [Diff<S>], context ctx: Int = 4) -> [Hunk] {
+  func prepending<S: StringProtocol>(_ prefix: String) -> (S) -> String {
+    return { prefix + $0 }
+  }
+
+  let (hunk, hunks) = diffs
+    .reduce((current: Hunk(), hunks: [Hunk]())) { cursor, diff in
+      var (current, hunks) = cursor
```
stephencelis: Please ignore the local mutation! 😬

PR #27, opened 2017-09-07
Refactor with a monoid

stephencelis:
Also discovered a lil issue with our failure message diff.

### DIFF
--- a/Sources/Diff/Diff.swift
+++ b/Sources/Diff/Diff.swift
@@ -44,16 +44,42 @@ private let plus = "+"
 private let figureSpace = "\u{2007}"
 
 public struct Hunk {
-  public fileprivate(set) var fstIdx: Int = 0
-  public fileprivate(set) var fstLen: Int = 0
-  public fileprivate(set) var sndIdx: Int = 0
-  public fileprivate(set) var sndLen: Int = 0
-  public fileprivate(set) var lines: [String] = []
+  public let fstIdx: Int
+  public let fstLen: Int
+  public let sndIdx: Int
+  public let sndLen: Int
+  public let lines: [String]
 
   public var patchMark: String {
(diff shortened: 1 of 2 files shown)


Sources/Diff/Diff.swift
```diff
@@ -44,70 +44,79 @@ private let plus = "+"
...
+        return (current + Hunk(sndLen: len, lines: diff.elements.map(prepending(plus))), hunks)
       }
-
-      return (current, hunks)
   }
 
-  return hunk.lines.contains(where: { $0.hasPrefix(minus) || $0.hasPrefix(plus) })
-    ? hunks + [hunk]
-    : hunks
+  return changed(hunk) ? hunks + [hunk] : hunks
```
stephencelis: Feel like there's a logical way to change the various `where` logic in the switch above to avoid this repetition, but haven't dug too deep.

Sources/SnapshotTesting/Diffable.swift
```diff
@@ -38,7 +38,10 @@ extension String: Diffable {
   public static func diffableDiff(_ fst: String, _ snd: String) -> (String, [XCTAttachment])? {
     guard fst != snd else { return nil }
 
-    let hunks = chunk(diff: diff(fst.split(separator: "\n"), snd.split(separator: "\n")))
+    let hunks = chunk(diff: diff(
+      fst.split(separator: "\n", omittingEmptySubsequences: false),
```
stephencelis: Surprising gotcha with the default behavior! You lose empty lines, even in the middle of a string!

``` swift
let str = """
Hello

And

Goodbye
"""

str.split(separator: "\n")
// ["Hello", "And", "Goodbye"]

str.split(separator: "\n").joined(separator: "\n") == str
// false
```

mbrandonw: we need tests!

PR #34, opened 2017-11-15
WebKit Snapshot Support

stephencelis:
Adds support for web page snapshots _via_ [`takeSnapshot(with:completionHandler:)`](https://developer.apple.com/documentation/webkit/wkwebview/2873260-takesnapshot).

Notes:

- [WebKit doesn't expose `WKSnapshotConfiguration.h` in its module map](http://www.openradar.me/33812691), so I've added an Objective-C shim to expose this method to Swift.
- CI's not quite there. I migrated Circle (naively) to 2.0, but haven't added Linux support there or figured out how to parallelize things. Both (Circle and Travis) seem to run on 10.2 instead of 10.3, so the WebKit API isn't available for macOS tests and I haven't figured out how to scope things yet.


.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
...
+          name: Generate Xcode project file
+          command: swift package generate-xcodeproj
+
+      - run:
+          name: Run iOS tests
+          command: xcodebuild test -scheme SnapshotTesting-Package -destination platform="iOS Simulator,name=iPhone 8,OS=11.0"
+          environment:
+            SNAPSHOT_ARTIFACTS: $CIRCLE_ARTIFACTS
+
+      # - run:
```
stephencelis: Disabled because Circle 2.0's macOS environment runs on 10.2, the WebKit APIs require 10.3, and I haven't figured out how to scope things properly with the shim.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 os:
   - linux
-  - osx
+  # - osx
```
stephencelis: Also running on 10.2.

Package.swift
```diff
@@ -2,6 +2,35 @@
 
 import PackageDescription
 
+let snapshotTestingDependencies: [Target.Dependency]
```
stephencelis: This file gets a whole lot messier. Suggestions on cleanup welcome!

README.md
```diff
@@ -2,9 +2,9 @@
 
 macOS [![CircleCI](https://circleci.com/gh/pointfreeco/swift-snapshot-testing.svg?style=svg)](https://circleci.com/gh/pointfreeco/swift-snapshot-testing) Linux [![Build Status](https://travis-ci.org/pointfreeco/swift-snapshot-testing.svg)](https://travis-ci.org/pointfreeco/swift-snapshot-testing)
 
-A library that records app data into test assertions. Because snapshot tests can capture the entirety of a data structure, they can cover far more surface area than a typical unit test.
+Automatically record app data into test assertions. Snapshot tests capture the entirety of a data structure and cover far more surface area than a typical unit test.
```
stephencelis: Just some minor copy changes/fixes here.

Sources/SnapshotTesting/UIKit.swift
```diff
@@ -63,8 +62,8 @@
   }
 
   extension UIImage: Snapshot {
-    public var snapshotFormat: Data {
-      return self.diffableData
+    public var snapshotFormat: UIImage {
+      return self
```
stephencelis: This ensures that images diff appropriately. The two-tiered protocol of `Diffable` and `Snapshot` is complicated! Would love to figure out a simpler way

Sources/WKSnapshotConfigurationShim/include/WKSnapshotConfigurationShim.h
```diff
@@ -0,0 +1,2 @@
+@import CoreGraphics;
+#import <WebKit/WKSnapshotConfiguration.h>
```
stephencelis: This shim should ideally be scoped for iOS 11 and macOS 10.3. Not quite sure how to organize everything yet.

stephencelis:
It's probably worth mentioning that this spike came out of getting image snapshot tests on our web repo! It required PostgreSQL and macOS, and this PR still only works on iOS since neither Circle nor Travis run on 10.3 yet 😞

Tests/SnapshotTestingTests/fixture.html
```diff
@@ -0,0 +1,6 @@
...
+(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
```
mbrandonw: lol maybe remove the google analytics from this!

Package.swift
mbrandonw: I just pushed this 223d567, maybe that's a lil better?

PR #35, opened 2017-11-16
Add buddybuild to take care of ios and macOS tests

mbrandonw:
This uses buddybuild to do our ios and macOS tests since they support high Sierra.


.circleci/config.yml
```diff
@@ -1,25 +0,0 @@
-version: 2
```
mbrandonw: so I removed circleci instead of travis cause I wasnt sure how to do linux testing with circleci 2.0 ... can you help with that? otherwise we could just use travis for linux testing?

PR #40, opened 2018-02-05
Xcode project for Carthage compatibility

### LINKED ISSUE #39: Carthage compatibility / Xcode project
Hi! 👋 

Would you accept a PR to add an `.xcodeproj`, making this library Carthage compatible?

mdiep:
Resolves #39.

I didn't add anything to the makefile to generate the Xcode project because the generated Xcode project isn't really suitable for distributing. (It's not composed well, missing required build settings, etc.)

Some iOS tests fail on my machine, but I think that's a function of which simulator you're using.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,14 @@
 .DS_Store
 /.build
 /Packages
-/*.xcodeproj
+
+# Xcode per-user config
+*.mode1
+*.mode1v3
+*.mode2v3
+*.perspective
+*.perspectivev3
+*.pbxuser
+xcuserdata
+*.xccheckout
+*.xcscmblueprint
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,13 @@ install:
   - if [ $TRAVIS_OS_NAME = linux ]; then
       eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)";
     fi
-script:
-  - swift test
+matrix:
+  include:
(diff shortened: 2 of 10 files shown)


stephencelis:
Nice!

> I didn't add anything to the makefile to generate the Xcode project because the generated Xcode project isn't really suitable for distributing. (It's not composed well, missing required build settings, etc.)

Ah, is this hand-rolled then? Is it possible to get in a shape where we can use `swift package generate-xcodeproj` with the `--xcconfig-overrides` option? _E.g._:

https://github.com/pointfreeco/swift-web/blob/dc1c5b700f698a16f7c736ea0f7d7bdc2e5d3d37/Makefile#L14

Or is there still some config that can't be done this way?

mdiep:
The generated project is set up to generate 3 frameworks:

- `Diff.framework`
- `WKSnapshotConfigurationShim.framework`
- `SnapshotTesting.framework`

I'm also not sure how multiple platforms work on the generated project.

The pain of having 3 frameworks makes it worth manually maintaining this IMO, but if you want it to be generated, I'll do it. ☺️

stephencelis:
Thoughts, @mbrandonw? Our workflow right now is a liberal sprinkling of `swift package generate-xcodeproj`, so wondering how this would change things.

mdiep:
_bump_

mbrandonw:
Hey! I'm a lil torn on this one. Considering our flow of working on this project, it is very likely we will accidentally blow away the xcodeproj changes on accident and break it for everyone.

But also, at the same time, we haven't made an official 0.1.0 release yet, so anyone using this is walking into a minefield already.

So, I guess let's merge and try it out!

Thanks @mdiep!

PR #42, opened 2018-04-01
Some fixes around WebKit snapshots

stephencelis:
This'll hopefully let us run screen shot tests on Circle!


Sources/Diff/Diff.swift
```diff
@@ -35,9 +35,10 @@ public func diff<A: Hashable>(_ fst: [A], _ snd: [A]) -> [Diff<A>] {
     let sndDiff = snd.isEmpty ? [] : [Diff(type: .second, elements: snd)]
     return fstDiff + sndDiff
   } else {
-    return diff(Array(fst.prefix(upTo: fstIdx)), Array(snd.prefix(upTo: sndIdx)))
```
stephencelis: This took awhile to compile, so breaking it up into multiple expressions.

stephencelis:
Passing on Circle. Weird stuff on Travis. Let's revisit since `master` is broken as is.

PR #44, opened 2018-04-17
Add attachments before calling XCTFail

creister:
This ensures attachments are added even if `XCTestCase.continueAfterFailure ==
false` (which is the default in UI Tests)


stephencelis:
Oh wow! Thanks for the PRs! Missed them somehow. Will review them later today, thanks!

stephencelis:
Thanks again!

PR #45, opened 2018-05-03
Use Device UIKit scale factor for loading images from file

creister:
This fixes screenshot comparisons failing on iPhone Plus models that use a
3x scale factor instead of the more common 2x scale factor


stephencelis:
Thanks!

PR #46, opened 2018-05-28
Use custom snapshot format

stephencelis:
There are enough changes to `dump` formatting in 4.2 to wanna own this a bit more and simplify things.


stephencelis:
Prelude diff: https://github.com/pointfreeco/swift-prelude/compare/custom-snapshot-format
Web diff: https://github.com/pointfreeco/swift-web/compare/custom-snapshot-format

Sources/SnapshotTesting/SnapshotTesting.swift
```diff
@@ -116,3 +108,57 @@ private var trackSnapshots = {
     print("Found \(count) stale snapshot\(count == 1 ? "" : "s"):\n\n\(list)")
   }
 }()
+
+public protocol SnapshotStringConvertible {
```
stephencelis: We now have a protocol that we can use to customize the snapshot of types. It takes precedence over `CustomStringConvertible` and `CustomDebugStringConvertible`.

Sources/SnapshotTesting/SnapshotTesting.swift
```diff
@@ -116,3 +108,57 @@ private var trackSnapshots = {
...
+  var snapshotDescription: String { get }
+}
+
+extension NSObject: SnapshotStringConvertible {
+  public var snapshotDescription: String {
+    return self.debugDescription
+      .replacingOccurrences(of: ": 0x[\\da-f]+", with: "", options: .regularExpression)
```
stephencelis: The logic for replacing `NSObject` pointer stuff now lives on `NSObject`.

Sources/SnapshotTesting/SnapshotTesting.swift
```diff
@@ -116,3 +108,57 @@ private var trackSnapshots = {
...
+    description = count == 1 ? "1 element" : "\(count) elements"
+  case (_, .dictionary?):
+    description = count == 1 ? "1 key/value pair" : "\(count) key/value pairs"
+  case (_, .set?):
+    description = count == 1 ? "1 member" : "\(count) members"
+  case (_, .tuple?):
+    description = count == 1 ? "(1 element)" : "(\(count) elements)"
+  case (_, .optional?):
+    let subjectType = String(describing: mirror.subjectType)
+      .replacingOccurrences(of: " #\\d+", with: "", options: .regularExpression)
```
stephencelis: In Swift 4.1, nested types sometimes become, _e.g._ `User #1` instead of `User`. This removes that. I thought maybe you could do something clever like use backticks to make a type name weird, but that's not allowed, so I think this is p bulletproof.

PR #63, opened 2018-10-18
Remove WKWebView overload and update image diffing

stephencelis:


Sources/SnapshotTesting/Strategy/NSImage.swift
```diff
@@ -117,4 +100,23 @@ private func context(for cgImage: CGImage) -> CGContext? {
   context.draw(cgImage, in: CGRect(x: 0, y: 0, width: cgImage.width, height: cgImage.height))
   return context
 }
+
+private func diff(_ old: NSImage, _ new: NSImage) -> NSImage {
+  let oldCiImage = CIImage(cgImage: old.cgImage(forProposedRect: nil, context: nil, hints: nil)!)
```
stephencelis: Using CoreImage now for diffing. It produces something very close to what we see in Kaleidoscope.

Sources/SnapshotTesting/Strategy/NSView.swift
```diff
@@ -8,17 +8,23 @@ extension Strategy {
   }
 
   public static func view(precision: Float) -> Strategy<NSView, NSImage> {
-    return Strategy.image(precision: precision).contramap {
-      precondition(!($0 is WKWebView), """
```
stephencelis: The overload made error messages really bad. So instead I've decided to bake the logic in here.

Sources/SnapshotTesting/Strategy/UIImage.swift
```diff
@@ -114,4 +96,19 @@ private func context(for cgImage: CGImage, data: UnsafeMutableRawPointer? = nil)
   context.draw(cgImage, in: CGRect(x: 0, y: 0, width: cgImage.width, height: cgImage.height))
   return context
 }
+
+private func diff(_ old: UIImage, _ new: UIImage) -> UIImage {
```
stephencelis: Really wanna explore a Core Image DSL...

Sources/SnapshotTesting/Strategy/NSView.swift
stephencelis: Basically, if you passed an `A` that doesn't have a default strategy to `assertSnapshot(matching:)` without naming an explicit strategy, the compiler assumed the overload instead, producing this error:

> 🛑 Cannot convert value of type 'A' to expected argument type 'WKWebView'

PR #65, opened 2018-10-21
WIP: Give dictionary keys a stable order in snapshots

finestructure:
This PR isn't quite working as intended as of yet but maybe it'll help outline the issue I'm trying to solve and with your guidance I hope I can polish this up.

I'm reading plist files for editing and I want to make sure edits are stable via snapshotting. What I've found is that simple reading the files at different times will make the snapshots fail, due to random order of dictionary keys.

For some reason this patch still seems to allow things to shuffle around so I'm guessing I'm not quite there yet :)

Also, another problem will be data payloads whose pointers will change between runs:

```
           ▿ value: Parameter
             ▿ data: 2056 bytes
               - count: 2056
−              ▿ pointer: 0x0000000104021a30
−                - pointerValue: 4362213936
```

This should probably be just masked out as

```
           ▿ value: Parameter
             ▿ data: 2056 bytes
               - count: 2056
−              ▿ pointer: <pointer address>
−                - pointerValue: <pointer value>
```

Or maybe this should be its own new `Any` strategy?


stephencelis:
Thanks for the PR! And thanks for dealing with some of our recent API churn :) We're looking to stabilize things soon.

Ideally a `plist` strategy using `PropertyListEncoder` would be a nice way of capturing such snaps, but looks like there's no `sortedKeys` option like there is on `JSONEncoder` :(

Any chance you can attach a test (passing or failing) to this PR? Would love to play around with things.

mbrandonw:
> Ideally a `plist` strategy using `PropertyListEncoder` would be a nice way of capturing such snaps, but looks like there's no `sortedKeys` option like there is on `JSONEncoder` :(

Stephen and I just tested this and it turns out that `PropertyListEncoder` does sort the keys, and so it is fully legitimate and stable to make a `plist` snapshot strategy. Wanna give that a shot @feinstruktur or we can?

> Also, another problem will be data payloads whose pointers will change between runs:

We have a `purgePointers` helper at the bottom of this file that we use to stripe `NSObject` pointers. Looks like we should do it in a few more spots. Would be nice to get some test coverage on that too.

finestructure:
I've added a test file and a test using the any strategy. Please note the following:

```
func testData(forResource resource: String, ofType type: String) throws -> Data {
    // For this method to find files, they need to be added to a "Copy Bundle Resource" build
    // phase for the test target in Xcode. This won't work with "swift test" invocation.
    // I'm not what the best way is to support both Xcode and SPM test runs.
```

Pretty sure circle ci will fall over due to this.

I've run the test a few times and unless I've been really lucky, the key issue is solved. The only failure diffs are seeing are now the likes of

```
 ▿ 166 bytes
   - count: 166
−  ▿ pointer: 0x0000000100a11bd0
−    - pointerValue: 4305525712
+  ▿ pointer: 0x00000001039483d0
+    - pointerValue: 4355032016
```

I'll try and sort that out with the `purgePointers` helper you mentioned, @mbrandonw . And move this to its own strategy perhaps.

Not sure why I got failures with my other files - I'll try this again with a larger one to see if was perhaps due to deeper nesting. Unfortunately didn't find more time to look into it further just now.

finestructure:
Actually, revering my sorting change does not cause key order to be mixed up.

The way this originally occurred was that in my project a parsed and persited plist file was sorted differently. I'm beginning to suspect that either the normal plist encoding part or some step in modifying it messes up the key order.

stephencelis:
@finestructure Looks like the main issue is that `SnapshotStringConvertible` is used to format the node but child nodes are still rendered. So in this case `UnsafePointer` is the culprit. I think we need to refine the design here to prevent children from rendering. Just need to figure out what that should look like. We'll have a fix for this soon, though! In the meantime, does @mbrandonw's `plist` strategy no #73 unblock you?

finestructure:
@stephencelis I've plugged the latest changes with #73 into my project directly and it works great. All plist snapshot tests are now stable and the snapshots import fine into the target application as well. Fantastic! Thanks for your help and for plugging in that change, much appreciated :)

I'll go ahead and close this PR as I don't there's anything useful in here at this point.

PR #73, opened 2018-11-02
Adds a plist strategy

mbrandonw:
Inspired by #65 I've added a plist snapshot strategy.


Sources/SnapshotTesting/Strategy/Codable.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-extension Strategy where A: Encodable {
+extension Strategy where A: Encodable, B == String {
```
mbrandonw: @stephencelis ok to further constraint to string?

Sources/SnapshotTesting/Strategy/Codable.swift
```diff
@@ -9,8 +9,24 @@ extension Strategy where A: Encodable {
   }
 
   public static func json(_ encoder: JSONEncoder) -> Strategy<A, String> {
-    return Strategy.lines.pullback { encodable in
+    var strategy = Strategy.lines.pullback { (encodable: A) in
       try! String(decoding: encoder.encode(encodable), as: UTF8.self)
     }
+    strategy.pathExtension = "json"
```
mbrandonw: seems we'd want json path extension right?

also we could bring in overture to make this nicer, but then feel like that complicates supporting cocoapods and Carthage.

mbrandonw:
I don't know why the NSView screenshot changed... it says the diff is 0 bytes and I can't see any diff between the images. maybe mojave?

Sources/SnapshotTesting/Strategy/Codable.swift
stephencelis: Yeah we should probably reconstrain everything. Makes the pullback more annoyingly verbose, but namespaces better.

Sources/SnapshotTesting/Strategy/Codable.swift
```diff
@@ -9,8 +9,26 @@ extension Strategy where A: Encodable {
...
+    return strategy
   }
+
+  #if !os(Linux)
```
finestructure: Out of curiosity - are you excluding linux because PropertyListEncoder is not available? I’ve worked around this in my project by pulling in a file from the stdlib. See 
https://finestructure.co/blog/2018/10/7/make-propertylistdecoder-available-on-linux

PR #75, opened 2018-11-02
Better View Strategy

### LINKED ISSUE #69: Improve view strategy
Right now we have separate UIView/NSView and WKWebView strategies, but really we should have a single UIView/NSView strategy that traverses the hierarchy and creates snapshots of views that aren't captured with `layer.render(in:)`, including:

- GLKView
- SCNView
- SKView
- WKWebView

stephencelis:
Fixes #69.

This PR merges all the view strategies into a single `view` strategy (per platform). It handles things like nested web views, Sprite Kit views, and Scene Kit views. Prob gonna add some more tests before mergin...


Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -111,7 +111,7 @@ open class SnapshotTestCase: XCTestCase {
 
       let diffMessage = self.diffTool
         .map { "\($0) \"\(snapshotFileUrl.path)\" \"\(failedSnapshotFileUrl.path)\"" }
-        ?? "@\(Diff.minus)\n\"\(failedSnapshotFileUrl.path)\"\n@\(Diff.plus)\n\"\(snapshotFileUrl.path)\""
+        ?? "@\(Diff.minus)\n\"\(snapshotFileUrl.path)\"\n@\(Diff.plus)\n\"\(failedSnapshotFileUrl.path)\""
```
stephencelis: These were in a counterintuitive order...

Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
 
-extension Strategy {
-  public static var any: Strategy<A, String> {
-    return Strategy.lines.pullback { snap($0) }
+extension Strategy where B == String {
```
stephencelis: I've gone through and constrained all the extensions. The pullbacks become a lil more verbose, but the benefit in reusing names seems worth it.

Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -26,8 +26,10 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
     let subjectType = String(describing: mirror.subjectType)
       .replacingOccurrences(of: " #\\d+", with: "", options: .regularExpression)
     description = count == 0 ? "\(subjectType).none" : "\(subjectType)"
-  case (let value as SnapshotStringConvertible, _):
+  case (let value as AnySnapshotStringConvertible, _) where type(of: value).renderChildren:
```
stephencelis: This also contains a fix for #65 where we can opt into omitting child mirror nodes.

Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -49,22 +51,41 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
   return lines.joined()
 }
 
-public protocol SnapshotStringConvertible {
+public protocol AnySnapshotStringConvertible {
```
stephencelis: Renamed this to be a lil more intuitive.

Sources/SnapshotTesting/Strategy/NSImage.swift
```diff
@@ -111,10 +111,7 @@ private func diff(_ old: NSImage, _ new: NSImage) -> NSImage {
     width: max(old.size.width, new.size.width),
     height: max(old.size.height, new.size.height)
   )
-  let differenceCiImage = differenceFilter.outputImage!
-  let invertFilter = CIFilter(name: "CIColorInvert")!
```
stephencelis: Inverting was making it harder to see diffs when images start rendering blank (the difference image looked exactly like the original).

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
+#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
```
stephencelis: Alright a single place for ALL the view logic.

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
...
+      )
+    }
+  }
+}
+
+extension UIView: DefaultDiffable {
+  public static let defaultStrategy: Strategy<UIView, UIImage> = .view
+}
+#endif
+
```
stephencelis: Everything after here is private implementation details.

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
...
+private typealias ImageView = NSImageView
+private typealias View = NSView
+#elseif os(iOS) || os(tvOS) || os(watchOS)
+import UIKit
+private typealias Image = UIImage
+private typealias ImageView = UIImageView
+private typealias View = UIView
+#endif
+
+private func addImagesForRenderedViews(_ view: View) -> [Async<View>] {
```
stephencelis: This traverses a view hierarchy, inserting and collecting image views in an asynchronous fashion (so they can be cleaned up and removed later).

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
...
+            callback(imageView)
+          }
+        }
+      ]
+    }
+    ?? view.subviews.flatMap(addImagesForRenderedViews)
+}
+
+fileprivate extension View {
+  var snapshot: Async<Image>? {
```
stephencelis: This is a mini "strategy" for grabbing an image of a view. Could probably be cleaned up using a protocol :wink:

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
...
+fileprivate final class ScaledWindow: NSWindow {
+  override var backingScaleFactor: CGFloat {
+    return 2
+  }
+}
+#endif
+#endif
+
+fileprivate extension Array {
+  func sequence<A>() -> Async<[A]> where Element == Async<A> {
```
stephencelis: Snuck a lil `sequence` in here!

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -0,0 +1,213 @@
...
+#if os(macOS)
+extension Strategy where A == NSView, B == NSImage {
+  public static var view: Strategy {
+    return .view(precision: 1)
+  }
+
+  public static func view(precision: Float) -> Strategy {
+    return Strategy<NSImage, NSImage>.image(precision: precision).asyncPullback { view in
+      return view.snapshot ?? Async { callback in
+        addImagesForRenderedViews(view).sequence().run { views in
```
stephencelis: `addImagesForRenderedViews` returns `[Async<View>]`, a bunch of image view renderings that have been inserted in the view hierarchy so we can remove them later. We call `sequence` to transform it into `Async<[View]>`, run it so it calls back when all of the image views have been inserted, and finally we can snapshot the view before removing the image views.

Sources/SnapshotTesting/Strategy/NSImage.swift
mbrandonw: oh interesting. we could also do an onion skin style 50% alpha on the two images.

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -1,12 +1,24 @@
...
+import SceneKit
+import SpriteKit
 import WebKit
+
+#if os(iOS)
+typealias View = UIView
 let platform = "ios"
 #elseif os(macOS)
-import WebKit
+typealias View = NSView
```
mbrandonw: I think this can be dropped since you have it in `SnapshotTesting` and you are doing a testable import?

stephencelis: Ah yup, those used to be private!

Sources/SnapshotTesting/Strategy/NSImage.swift
stephencelis: Yup, thinkin we can probably refine the diff image in a number of ways.

PR #76, opened 2018-11-05
Add plist encoder for linux, allow plist snapshot testing on linux

finestructure:
Add PropertyList en/decoder for Linux to support plist snapshotting on Linux.

NB: `make test-linux` passes but `test-macos` and `make test-ios` fail - but they fail on master as well.

I'd love to be able to run my snapshot tests on my linux based CI as well but it would require this file `PropertyListEncoder.swift` to be included. (As I mentioned in a comment elsewhere, see https://finestructure.co/blog/2018/10/7/make-propertylistdecoder-available-on-linux for details.)

Perhaps not ideal to include this file but hopefully the stdlib will soon support that on Linux.

Alternatively, maybe I should put this in a small SPM dependency to be included?


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM swift:4.0
+FROM swift:4.2
```
finestructure: I had to lift this up or `PropertyListEncoder.swift` would not compile.

stephencelis:
Thanks for the PR! Including this sounds good to me while we wait for Apple to do so.

@mbrandonw Maybe we should come up with a plan on how to modularize these strategies as we add more and more.

mbrandonw:
@finestructure: @stephencelis and I chatted about this, and we think this makes most sense to split out into another library, something like `swift-snapshot-plist`. It would depend on this library and just provide the `plist` strategy. We want to do this so that if/when Apple supports plist on linux users of this library can just drop the `swift-snapshot-plist` dependency and get things working rather than having to wait on us to update `swift-snapshot-testing`.

I'll get that going, and then maybe you could re-open this PR off of that repo once its ready?

finestructure:
Sure! Just one thing I'd like to note: In my project I'm actually adding the plist encoder/decoder implementation without any `#if os(Linux)` guards. In other words I'm using it even on macOS, shadowing the stdlib's implementation. (No particular reason, it's just that it only occurred to me to do so when adding it to this PR.)

The point being that there should be no issue with respect to plist support shipping on Linux while the project is not updated to take advantage of it and shipping its own implementation for a while longer.

But it sounds like the plist encoder file is not the only reason for moving this to a separate lib – you're looking for more modularity as well, right? I.e. you wouldn't move it back once plist decoding ships for Linux, in which case it'd probably be more hassle than it's worth.

I'll go ahead and close this PR – just let me know if and where you'd find it useful and I'll be back ;)

mbrandonw:
> The point being that there should be no issue with respect to plist support shipping on Linux while the project is not updated to take advantage of it and shipping its own implementation for a while longer.

Ah interesting, if that's the case then I think we are fine bringing this in for now. We do think it makes sense to modularize the strategies at some point but we still have a bit more work we won't to do before we get there.

Stephen and I will chat more today about how to move forward!

stephencelis:
Let's get this merged in! I think we can safely remove the `#if os(Linux)` guard, too. Whenever Linux Swift ships with `PropertyListEncoder` we can just safely remove it all at once.

Sources/SnapshotTesting/PropertyListEncoder.swift
```diff
@@ -0,0 +1,1827 @@
+#if os(Linux)
```
stephencelis: Tried making a suggested change but GitHub doesn't seem to let me for new files. Can you remove this guard?

stephencelis:
Actually I was able to resolve the conflicts and make the changes myself. Will be merging in shortly. Thanks again for the PR, @finestructure!

PR #77, opened 2018-11-05
Use XcodeGen

stephencelis:
Unfortunately, SwiftPM has a few Xcode project generation issues that make it far from ideal to use.

- Xcode project generation is not idempotent: it will always generate different values on repeat runs, and it will always modify the current project in a way that leaves artifacts behind
- The project generates a shared scheme that prevents us from supporting Carthage

This PR switches things up to use XcodeGen instead!


stephencelis:
(Will look into CI failures. Tests work fine locally.)

Development.xcconfig
```diff
@@ -1 +0,0 @@
-OTHER_SWIFT_FLAGS=$(inherited) -Xfrontend -warn-long-expression-type-checking=80
```
stephencelis: Not bothering with this anymore.

Package.swift
```diff
@@ -2,30 +2,6 @@
...
 import PackageDescription
 
-#if os(Linux)
-let isLinux = true
-#else
-let isLinux = false
-#endif
-
-let shimTarget = Target.target(
-  name: "WKSnapshotConfigurationShim",
```
stephencelis: Dropping the shim. Doesn't appear to be necessary and complicates things.

Sources/SnapshotTesting/Internal.swift
```diff
@@ -3,7 +3,7 @@ import Cocoa
 typealias Image = NSImage
 typealias ImageView = NSImageView
 typealias View = NSView
-#elseif os(iOS) || os(tvOS) || os(watchOS)
```
stephencelis: watchOS doesn't include XCTest or unit testing in general, so no need to support!

Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -1,4 +1,3 @@
-import Diff
```
stephencelis: I moved `Diff` into the main lib to make things a bit easier.

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -104,42 +106,48 @@ private func addImagesForRenderedViews(_ view: View) -> [Async<View>] {
 
 fileprivate extension View {
   var snapshot: Async<Image>? {
-    #if os(iOS) || os(tvOS) || os(watchOS)
+    func inWindow<T>(_ perform: () -> T) -> T {
```
stephencelis: This is required for views that snapshot themselves in order to "trick" it into rendering @2x. This logic used to be hard-coded to WKWebViews but I noticed that SceneKit/SpriteKit views were also rendering the wrong size.

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -17,11 +20,10 @@ extension Strategy where A == NSView, B == NSImage {
 
   public static func view(precision: Float) -> Strategy {
     return Strategy<NSImage, NSImage>.image(precision: precision).asyncPullback { view in
-      return view.snapshot ?? Async { callback in
+      view.snapshot ?? Async { callback in
         addImagesForRenderedViews(view).sequence().run { views in
           let image = NSImage(data: view.dataWithPDF(inside: view.bounds))!
-          let scale = NSScreen.main!.backingScaleFactor
```
stephencelis: Plugging into a 1x screen made me realize that tests could fail.

project.yml
```diff
@@ -0,0 +1,28 @@
+name: SnapshotTesting
```
stephencelis: Here's the main project.yml file. Wasn't so bad once I saw the `$platform` trick!

PR #79, opened 2018-11-06
Don't require a test case subclass

### LINKED ISSUE #71: [Question] Rearchitect to not require subclassing SnapshotTestCase?
We use [Quick](https://github.com/Quick/Quick) for testing and are therefore required to subclass `QuickSpec`, which precludes us from also subclassing `SnapshotTestCase`. Is it possible to architect this as a set of extensions to `XCTestCase` so that it can be used alongside other testing libraries that are less flexible?

I would be happy to put some work into this and put up a PR if you think it would be possible/valuable.

stephencelis:
Fixes #71.

Hoping that a serial queue makes things reliable.


Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -1,126 +1,134 @@
...
+    try fileManager.createDirectory(at: snapshotDirectoryUrl, withIntermediateDirectories: true)
+
+    let tookSnapshot = XCTestExpectation(description: "Took snapshot")
+    var optionalDiffable: B?
+    strategy.snapshotToDiffable(value).run { b in
+      optionalDiffable = b
+      tookSnapshot.fulfill()
+    }
+    let result = XCTWaiter.wait(for: [tookSnapshot], timeout: timeout)
+    switch result {
```
stephencelis: We need to manually fail based on the `XCTWaiter` result now.

Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -1,126 +1,134 @@
...
-      try fileManager.createDirectory(at: snapshotDirectoryUrl, withIntermediateDirectories: true)
-
-      let tookSnapshot = self.expectation(description: "Took snapshot")
-      var optionalDiffable: B?
-      strategy.snapshotToDiffable(value).run { b in
-        optionalDiffable = b
-        tookSnapshot.fulfill()
-      }
-      #if os(Linux)
-      self.waitForExpectations(timeout: timeout)
```
stephencelis: I'm hoping this doesn't fail. Linux got the additional APIs in recent snapshots...

stephencelis: Dang, merged less than a week ago: https://github.com/apple/swift-corelibs-xctest/pull/228

Guess we can't merge and support Linux unless we lock Linux to a recent snapshot.

Killectro:
Bless you both @stephencelis @mbrandonw, very excited to give this a test!

PR #80, opened 2018-11-06
assertSnapshot(of:matching:) 🔜 (matching:as:)

stephencelis:
Reads a bit nicer.

### DIFF
--- a/Tests/SnapshotTestingTests/SnapshotTestingTests.swift
+++ b/Tests/SnapshotTestingTests/SnapshotTestingTests.swift
@@ -29,19 +29,19 @@ class SnapshotTestingTests: SnapshotTestCase {
   func testWithAny() {
     struct User { let id: Int, name: String, bio: String }
     let user = User(id: 1, name: "Blobby", bio: "Blobbed around the world.")
-    assertSnapshot(matching: user, as: .any)
-    assertSnapshot(matching: Data("Hello, world!".utf8), as: .any)
-    assertSnapshot(matching: URL(string: "https://www.pointfree.co")!, as: .any)
+    assertSnapshot(matching: user, as: .dump)
+    assertSnapshot(matching: Data("Hello, world!".utf8), as: .dump)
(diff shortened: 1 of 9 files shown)


Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -29,48 +29,48 @@ class SnapshotTestingTests: SnapshotTestCase {
   func testWithAny() {
     struct User { let id: Int, name: String, bio: String }
     let user = User(id: 1, name: "Blobby", bio: "Blobbed around the world.")
-    assertSnapshot(of: .any, matching: user)
-    assertSnapshot(of: .any, matching: Data("Hello, world!".utf8))
-    assertSnapshot(of: .any, matching: URL(string: "https://www.pointfree.co")!)
+    assertSnapshot(matching: user, as: .any)
```
stephencelis: Thinking we may wanna rename `.any` to `.dump`

``` swift
assertSnapshot(matching: user, as: .dump)
```

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -188,9 +196,9 @@ class SnapshotTestingTests: SnapshotTestCase {
     #endif
     if #available(macOS 10.14, *) {
       label.text = "Hello."
-      assertSnapshot(of: .view(precision: 0.9), matching: label, named: platform)
+      assertSnapshot(matching: label, as: .view(precision: 0.9), named: platform)
```
stephencelis: Think `.view` and `.viewController` and `.scene` should all probably be `.image`:

``` swift
assertSnapshot(matching: scene, as: .image(size: .init(width: 50, height: 50)))
assertSnapshot(matching: label, as: .image(precision: 0.9))
```

Or `.rendered`:

``` swift
assertSnapshot(matching: scene, as: .rendered(size: .init(width: 50, height: 50)))
assertSnapshot(matching: label, as: .rendered(precision: 0.9))
```

PR #81, opened 2018-11-07
Improve cocoapods support

cadizjavier:
This PR includes 2 changes:
1 - When using the library with Cocoapods i get the following error:
`Cannot load underlying module for 'XCTest'`
A fix in the `SnapshotTesting.podspec` file was added to include the `XCTest` framework as a dependency.

2- Minor improvement in the README to make more explicit the support for Cocoapods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ This library should be considered alpha, and not stable. Breaking changes will h
 
 ## Installation
 
+### Swift Package Manager
+
 ```swift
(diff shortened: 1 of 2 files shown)


stephencelis:
Oh interesting. This change was needed for XcodeGen for a similar reason:

https://github.com/pointfreeco/swift-snapshot-testing/blob/master/project.yml#L11

Thanks for the PR!

PR #82, opened 2018-11-07
Renaming the generics and default protocol.

mbrandonw:


Sources/SnapshotTesting/Strategy/SceneKit.swift
```diff
@@ -7,20 +7,20 @@ import UIKit
...
-extension Strategy where A == SCNScene, B == UIImage {
+extension Strategy where Snapshottable == SCNScene, Format == UIImage {
   public static func image(size: CGSize, precision: Float = 1) -> Strategy {
     return .scnScene(size: size, precision: precision)
   }
 }
 #endif
 
-fileprivate extension Strategy where A == SCNScene, B == Image {
+fileprivate extension Strategy where Snapshottable == SCNScene, Format == Image {
```
mbrandonw: this reads pretty nicely now

Issue #70, opened 2018-10-25
Add size configuration option to view snapshots


stephencelis:
I'd be nice to configure the view size and it'd be nice to have some easily configured options, like device sizes and screen resolutions.

Responsive design mode has some nice options:

- [ ] iPhone SE (320x568)
- [ ] iPhone 8 (375x667)
- [ ] iPhone 8 Plus (414x736) (3x size...can we fake it?)
- [ ] iPad mini (7.9") / iPad (9.7") (1024x768)
- [ ] iPad Pro (10.5") (1112x834)
- [ ] iPad Pro (12.9") (1366x1024)
- [ ] 800x600
- [ ] 1366x768
- [ ] 1920x1080

It'd also be nice to adapt this to have an "below the fold" option, which renders with an adaptive height and configurable fold rendering option.

Issue #83, opened 2018-11-07
Add fatal error for views that have zero width or height


stephencelis:
The current error isn't so helpful.

PR #85, opened 2018-11-08
Sizing Stuff

stephencelis:
Some fun things to think about. I dunno if this is the right API but I think it's worth starting a discussion!


Sources/SnapshotTesting/Strategy/CGSize.swift
```diff
@@ -0,0 +1,88 @@
...
+import CoreGraphics
+#endif
+
+public enum Orientation {
+  case horizontal
+  case vertical
+}
+
+extension CGSize {
+  static func iPhoneSe(_ orientation: Orientation) -> CGSize {
```
stephencelis: Minor thing: do we wanna do `iPhoneSE` instead? I'm keen on sticking with our camelcasing guns.

Sources/SnapshotTesting/Strategy/NSViewController.swift
```diff
@@ -7,6 +7,14 @@ extension Strategy where Snapshottable == NSViewController, Format == NSImage {
   }
 
   public static func image(precision: Float) -> Strategy {
+    return .image(precision: precision, size: nil)
+  }
+
+  public static func image(precision: Float = 1, size: CGSize) -> Strategy {
```
stephencelis: Just some basic overloads for taking `precision`, `size`, or both, while delegating to a nullable, private implementation.

Tests/SnapshotTestingTests/SnapshotTestCaseTests.swift
```diff
@@ -62,6 +62,13 @@ class SnapshotTestCaseTests: TestCase {
     let view = UIButton(type: .contactAdd)
     assertSnapshot(matching: view)
     assertSnapshot(matching: view, as: .recursiveDescription)
+
+    assertSnapshot(matching: view, as: .image(size: .iPadMini))
```
stephencelis: Examples in practice.

stephencelis:
One problem with having everything as simple statics on `CGSize` is that this API wouldn't be so reusable for `UIViewController` where `UITraitCollection` comes into play. Dunno if you have ideas on how `UITraitCollection` may be used in the future and how it may affect this design.

Sources/SnapshotTesting/Strategy/CGSize.swift
stephencelis: (I could remove this file if we can't figure it out and the other changes should hopefully still be useful!)

stephencelis:
(This fixes #70 and #83.)

PR #90, opened 2018-11-09
Add trait collection strategies

stephencelis:
Mostly using code we developed earlier.


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
+# Contributor Covenant Code of Conduct
```
stephencelis: Copied this over from our other repos. I can make it its own PR...

Tests/SnapshotTestingTests/SnapshotTestCaseTests.swift
```diff
@@ -191,6 +191,86 @@ class SnapshotTestCaseTests: TestCase {
...
+
+  func testTraits() {
+    #if os(iOS)
+    class MyViewController: UIViewController {
+      let label = UILabel()
+
+      override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.label.text = "What's the point?"
```
mbrandonw: lol nice

Sources/SnapshotTesting/Strategy/UIViewController.swift
```diff
@@ -28,4 +54,166 @@ extension Strategy where Snapshottable == UIViewController, Format == String {
...
+  parent.view.frame.size = size
+  parent.preferredContentSize = parent.view.frame.size
+  parent.addChild(viewController)
+  parent.view.addSubview(viewController.view)
+  parent.setOverrideTraitCollection(traits, forChild: viewController)
+
+  viewController.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+  viewController.view.frame = parent.view.frame
+
+  parent.beginAppearanceTransition(true, animated: false)
```
stephencelis: Should we only do this conditionally? Does it matter?

Sources/SnapshotTesting/Strategy/UIViewController.swift
```diff
@@ -28,4 +54,166 @@ extension Strategy where Snapshottable == UIViewController, Format == String {
...
+
+  public static let iPadPro12_9 = Environment(device: .iPadPro12_9, orientation: .landscape)
+
+  public static func iPadPro12_9(_ orientation: Orientation) -> Environment {
+    return Environment(device: .iPadPro12_9, orientation: orientation)
+  }
+}
+
+private enum Device {
+  case iPhoneSe
```
mbrandonw: what about an alternate design of statics vars/funcs on `UITraitCollection` for the various device configurations? think that could work?

stephencelis: I thought about that but what about size?

mbrandonw: maybe `Environment` is `struct { traits: [UITraitCollection], size: CGSize }` with statics? then people could make their own, like a tv or iPhone X?

mbrandonw: also could we make a notch pullback? 😛 `notchramap`

stephencelis: Ah cool, I like that! Not sure what's required to make iPhone X really work since the safe area insets are different. I've tried before but it wasn't quite right.

Sources/SnapshotTesting/Strategy/UIViewController.swift
```diff
@@ -28,4 +51,205 @@ extension Strategy where Snapshottable == UIViewController, Format == String {
...
+  public let traits: UITraitCollection
+
+  public init(size: CGSize, traits: UITraitCollection) {
+    self.size = size
+    self.traits = traits
+  }
+
+  public static let iPhoneSe = Environment.iPhoneSe(.portrait)
+
+  public static func iPhoneSe(_ orientation: Orientation) -> Environment {
```
stephencelis: It's a lil more verbose. We could create helpers but not sure it's worth it.

PR #92, opened 2018-11-10
UIView/Controller Strategy Improvements

### LINKED ISSUE #78: Support UIAppearance and visual effect views
iOSSnapshotTestCase supports UIAppearance and visual effect views using a window and `drawViewHierarchyInRect`:

https://github.com/uber/ios-snapshot-test-case/blob/de1c714952ca6b7f1a40670135297fd761ba43cf/FBSnapshotTestCase/Categories/UIImage%2BSnapshot.m#L39-L70

We typically write tests against framework, not app, targets, where I'm not sure this is even possible. We should investigate support, though, maybe at least by introspecting whether or not there's a window we can use.

stephencelis:
Some changes around snapshotting UIView/Controllers.

Fixes #78.

### DIFF
--- a/Sources/SnapshotTesting/Strategy/View.swift
+++ b/Sources/SnapshotTesting/Strategy/View.swift
@@ -75,13 +75,13 @@ extension Strategy where Snapshottable == UIView, Format == UIImage {
 
   static func image(precision: Float, size: CGSize?) -> Strategy {
     return SimpleStrategy.image(precision: precision).asyncPullback { view in
-      let initialFrame = view.frame
-      view.frame.origin = CGPoint(x: .max, y: .max)
+      let initialFrame = view.frame                                                                     
       if let size = size { view.frame.size = size }
       guard view.frame.width > 0, view.frame.height > 0 else {
         fatalError("View not renderable to image at size \(view.frame.size)")
       }
       if let window = UIApplication.shared.keyWindow, window != view {
+        view.frame.origin = CGPoint(x: .max, y: .max)
         window.addSubview(view)
       }
       view.setNeedsLayout()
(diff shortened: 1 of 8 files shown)


Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -75,17 +75,35 @@ extension Strategy where Snapshottable == UIView, Format == UIImage {
...
   static func image(precision: Float, size: CGSize?) -> Strategy {
     return SimpleStrategy.image(precision: precision).asyncPullback { view in
-      let initialSize = view.frame.size
+      let initialFrame = view.frame                                                                     
       if let size = size { view.frame.size = size }
       guard view.frame.width > 0, view.frame.height > 0 else {
         fatalError("View not renderable to image at size \(view.frame.size)")
       }
+      if let window = UIApplication.shared.keyWindow, window != view {
+        view.frame.origin = CGPoint(x: .max, y: .max)
```
stephencelis: When the origin is `.zero` views layout very strangely because of the window some kind of margins/safe-area influence. Not sure if there's a better way of handling this.

Sources/SnapshotTesting/Strategy/View.swift
```diff
@@ -62,34 +62,58 @@ extension NSView: DefaultSnapshottable {
...
+          parent.view.frame.origin = CGPoint(x: .max, y: .max)
+          window.addSubview(parent.view)
+        }
+        parent.view.setNeedsLayout()
+        parent.view.layoutIfNeeded()
+        return parent.view.snapshot ?? Async { callback in
+          addImagesForRenderedViews(parent.view).sequence().run { views in
+            let cleanup = {
+              views.forEach { $0.removeFromSuperview() }
+              view.frame = initialFrame
```
mbrandonw: can this be a `defer`? not sure I'm fully following this code correctly.

stephencelis: unfortunately not since one of the paths happens in a callback :(

PR #96, opened 2018-11-15
Fix UIImage Comparison Edge Case

stephencelis:
Every once in awhile the new UIImage can't produce a context, etc.


Sources/SnapshotTesting/Strategy/UIImage.swift
```diff
@@ -58,10 +58,10 @@ private func compare(_ old: UIImage, _ new: UIImage, precision: Float) -> Bool {
...
   var oldBytes = [UInt8](repeating: 0, count: byteCount)
   guard let oldContext = context(for: oldCgImage, data: &oldBytes) else { return false }
-  guard let newContext = context(for: newCgImage) else { return false }
   guard let oldData = oldContext.data else { return false }
-  guard let newData = newContext.data else { return false }
-  if memcmp(oldData, newData, byteCount) == 0 { return true }
+  if let newContext = context(for: newCgImage), let newData = newContext.data {
+    if memcmp(oldData, newData, byteCount) == 0 { return true }
+  }
   let newer = UIImage(data: new.pngData()!)!
```
stephencelis: Just a quick way to avoid this work when we can.

PR #97, opened 2018-11-16
Any Improvements

stephencelis:
Few small improvements to the `dump` strategy.

### DIFF
--- a/Sources/SnapshotTesting/Strategy/Any.swift
+++ b/Sources/SnapshotTesting/Strategy/Any.swift
@@ -73,9 +73,15 @@ extension Date: AnySnapshotStringConvertible {
 }
 
 extension NSObject: AnySnapshotStringConvertible {
+  #if canImport(ObjectiveC)
   @objc open var snapshotDescription: String {
     return purgePointers(self.debugDescription)
   }
+  #else
+  open var snapshotDescription: String {
+    return purgePointers(self.debugDescription)
+  }
+  #endif
 }
 
 extension String: AnySnapshotStringConvertible {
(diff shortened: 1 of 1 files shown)


Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -41,8 +39,8 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
     let subjectType = String(describing: mirror.subjectType)
       .replacingOccurrences(of: " #\\d+", with: "", options: .regularExpression)
     description = count == 0 ? "\(subjectType).\(value)" : "\(subjectType)"
-  default:
-    description = "(indescribable)"
+  case (let value, _):
+    description = String(describing: value)
```
stephencelis: Seems like a nice default over `(indescribable)`.

Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -75,7 +73,7 @@ extension Date: AnySnapshotStringConvertible {
 }
 
 extension NSObject: AnySnapshotStringConvertible {
-  public var snapshotDescription: String {
+  @objc open var snapshotDescription: String {
```
stephencelis: Required for overriding in other modules.

Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -30,8 +30,6 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
     description = value.snapshotDescription
   case (let value as AnySnapshotStringConvertible, _):
     return "\(indentation)- \(name.map { "\($0): " } ?? "")\(value.snapshotDescription)\n"
-  case (let value as CustomDebugStringConvertible, _):
```
stephencelis: I think `CustomDebugStringConvertible` assumes a lot, and usually embeds the descriptions of other types, bailing out of things like `AnySnapshotStringConvertible`, so less of a good idea.

PR #98, opened 2018-11-16
Allocate enough bytes for image comparison

mstultz:
### What

Fixes a crash I was seeing where `oldCgImage.bytesPerRow` did not match `newCgImage.bytesPerRow`. The buffer is reused for both images, so it should allocate based on the larger value of the two.


mbrandonw:
nice!

PR #103, opened 2018-11-20
Snapshot dictionaries and sets deterministically

### LINKED ISSUE #101: Dictionary snapshots
Dictionary snapshotting is not deterministic. Let's consider the following snapshot.

```
let dictionary = ["first": 1, "second": 2, "third": 3]
assertSnapshot(matching: dictionary)
```

Each time the snapshot is made, the snapshot result is different.

```
▿ 3 key/value pairs
  ▿ (2 elements)
    - key: "third"
    - value: 3
  ▿ (2 elements)
    - key: "first"
    - value: 1
  ▿ (2 elements)
    - key: "second"
    - value: 2
```

```
▿ 3 key/value pairs
  ▿ (2 elements)
    - key: "first"
    - value: 1
  ▿ (2 elements)
    - key: "third"
    - value: 3
  ▿ (2 elements)
    - key: "second"
    - value: 2
```

Is there any way to handle this scenario correctly? Should we have a strategy for `Dictionary<Comparable, T>` so the snapshot is captured by sorting the keys in the first place?

Thanks

luisrecuenco:
Fixes #101 by sorting the children for sets and dictionaries recursively.


Sources/SnapshotTesting/Strategy/Any.swift
```diff
@@ -35,6 +38,7 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
   case (_, .class?), (_, .struct?):
     description = String(describing: mirror.subjectType)
       .replacingOccurrences(of: " #\\d+", with: "", options: .regularExpression)
+    children = sort(children)
```
luisrecuenco: I was doubtful at first to introduce this, as it will force people regenerate a lot of snapshots, but I think that, even if member-wise inits change when the order of the fields changes, the snapshot content shouldn't be aware of this, as the content is really the same. What do you think?

stephencelis: Ha, at first thought I was doubtful but after thinking it through I'm coming around to it. Let's do it for now and we can always revisit if there are any consequences.

PR #104, opened 2018-11-21
Disable bitcode for iOS framework

luisrecuenco:
This pull request aims at fixing the following problem that arises when trying to build for arm64 architecture when using Carthage, which always builds fat binaries.

`ld: warning: Could not find auto-linked framework 'XCTest'`

Module linking to XCTest assumes bitcode dependencies, but XCTest doesn't actually support bitcode.

### DIFF
--- a/SnapshotTesting.xcodeproj/project.pbxproj
+++ b/SnapshotTesting.xcodeproj/project.pbxproj
@@ -629,6 +629,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				ENABLE_BITCODE = NO;
 				FRAMEWORK_SEARCH_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/Library/Frameworks";
 				INFOPLIST_FILE = Sources/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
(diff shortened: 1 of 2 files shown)


stephencelis:
Hi @luisrecuenco! We generate our project file automatically using [XcodeGen](https://github.com/yonaskolb/XcodeGen). Can you modify the [project.yml](https://github.com/pointfreeco/swift-snapshot-testing/blob/010afed17a6680936f24f69f6e823cbbc69acaea/project.yml#L14-L15) file and re-run `xcodegen` instead?

luisrecuenco:
I don't see anything related to bitcode [here](https://github.com/yonaskolb/XcodeGen/blob/master/Docs/ProjectSpec.md). Does XcodeGen support it?

stephencelis:
You can add any arbitrary setting under the `settings:` node. I think the following will do:

``` yml
settings:
  ENABLE_BITCODE: 'NO'
```

luisrecuenco:
Wow, that was super easy! 😄 But I'm not finding as easy to just disable bitcode for a specific platform, in this case, for iOS. I guess it's just my XcodeGen ignorance.

stephencelis:
Given that it needs to be disabled on iOS and tvOS, I think it's safe to disable it for all platforms. Don't see the harm in macOS ignoring the setting when it builds. I did a lil test run locally and was able to disable it by adding `ENABLE_BITCODE: false` directly above the `FRAMEWORK_SEARCH_PATHS` setting. Wanna make that change and I can merge?

luisrecuenco:
I’m afraid bitcode must be enabled for tvOS, right?

https://help.apple.com/xcode/mac/current/#/devbbdc5ce4f

stephencelis:
For submission to the app store, yes, but not for testing, right?

luisrecuenco:
I guess you are right, but I wonder how XcodeGen allows to disable bitcode
just for one platform for other libraries.

In any case, I’ll change the XcodeGen file right away. Thanks!

stephencelis:
I think we'd have to unroll the config file and define each product/platform manually, unfortunately, though I don't have much more XcodeGen experience than you 😅

luisrecuenco:
Done. The truth is that the `ENABLE_BITCODE` build option only exists for iOS. What XcodeGen does for macOS and tvOS is simply add it as User-Defined setting, which is not correct per se, but I guess it's not harm.

PR #105, opened 2018-11-26
Render with trait collection

stephencelis:
Went down a bit of a rabbit hole today trying to get iPhone X support and discovered some things along the way. This is still a WIP but will get around to tvOS soon!


Sources/SnapshotTesting/Strategy/CALayer.swift
```diff
@@ -28,16 +28,29 @@ import UIKit
...
+        layer.image(for: traits)
       }
+  }
+}
+
+extension CALayer {
+  func image(for traits: UITraitCollection) -> UIImage {
+    let renderer: UIGraphicsImageRenderer
+    if #available(iOS 11.0, *) {
+      renderer = UIGraphicsImageRenderer(size: self.bounds.size, format: .init(for: traits))
```
stephencelis: This allows the renderer to take things like gamut/scale from the trait collection to render a proper image size on simulators that may differ!

Sources/SnapshotTesting/Strategy/UIViewController.swift
```diff
@@ -7,36 +7,21 @@ extension Strategy where Snapshottable == UIViewController, Format == UIImage {
   }
 
   public static func image(
-    drawingHierarchyInKeyWindow: Bool = false,
```
stephencelis: Removing this option for now. Will try to add back soon but it was making the refactor a bit of a pain.

stephencelis: (And not sure what to do about the fact that we need to use a custom window for safe area insets. Maybe you have ideas?)

Tests/SnapshotTestingTests/SnapshotTestCaseTests.swift
```diff
@@ -194,82 +198,121 @@ class SnapshotTestCaseTests: TestCase {
...
+          self.trailingLabel.translatesAutoresizingMaskIntoConstraints = false
+          self.bottomLabel.translatesAutoresizingMaskIntoConstraints = false
+
+          self.view.addSubview(self.topLabel)
+          self.view.addSubview(self.leadingLabel)
+          self.view.addSubview(self.trailingLabel)
+          self.view.addSubview(self.bottomLabel)
+
+          NSLayoutConstraint.activate([
+            self.topLabel.topAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.topAnchor),
```
stephencelis: Updated to demonstrate safe area insets working.

Tests/SnapshotTestingTests/SnapshotTestCaseTests.swift
```diff
@@ -194,82 +198,121 @@ class SnapshotTestCaseTests: TestCase {
...
+            self.topLabel.centerXAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.centerXAnchor),
+            self.leadingLabel.leadingAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.leadingAnchor),
+            self.leadingLabel.trailingAnchor.constraint(lessThanOrEqualTo: self.view.safeAreaLayoutGuide.centerXAnchor),
+//            self.leadingLabel.trailingAnchor.constraint(equalTo: self.view.safeAreaLayoutGuide.centerXAnchor),
```
stephencelis: This "fixes" a test that failed with trait collection override of 3x scale.

Tests/SnapshotTestingTests/SnapshotTestCaseTests.swift
```diff
@@ -194,82 +198,121 @@ class SnapshotTestCaseTests: TestCase {
...
-      override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
-
-        self.label.font = .preferredFont(forTextStyle: .body, compatibleWith: self.traitCollection)
+      let viewController = MyViewController()
+
+      assertSnapshot(matching: viewController, as: .image(on: .iPhoneSe), named: "iphone-se")
+      assertSnapshot(matching: viewController, as: .image(on: .iPhone8), named: "iphone-8")
+      assertSnapshot(matching: viewController, as: .image(on: .iPhone8Plus), named: "iphone-8-plus")
+      assertSnapshot(matching: viewController, as: .image(on: .iPhoneX), named: "iphone-x")
```
stephencelis: Now supporting iPhone X, XR, and XS Max!

stephencelis:
@mbrandonw This is good to go I think. I can go back and add some static trait collections for tvOS later.

mluisbrown:
> a completely simulator-agnostic rendering environment

That would be awesome if possible. Currently we're having to force the use of the iPhone SE (or other 2x scale device) for UI snapshot testing (FBSnapshotTestCase has the same issue btw).

.circleci/config.yml
```diff
@@ -15,14 +15,10 @@ jobs:
       - run:
           name: Run iOS tests
           command: make test-ios
-          environment:
-            SNAPSHOT_ARTIFACTS: /tmp/__SnapshotArtifacts__
```
stephencelis: This wasn't working with Circle as expected for `xcodebuild`. Had to add to the schemes via XcodeGen.

Sources/SnapshotTesting/Strategy/CALayer.swift
```diff
@@ -28,16 +28,30 @@ import UIKit
 
 extension Strategy where Snapshottable == CALayer, Format == UIImage {
   public static var image: Strategy {
-    return .image(precision: 1)
```
stephencelis: Rather than maintain a jumble of overloads, paring it down to one property and one function with all defaults.

Sources/SnapshotTesting/Strategy/CALayer.swift
stephencelis: Edit: I've removed the gamut/scale from the device trait collections for now, though people are still free to use them. I was seeing issues with layout and console warnings with CGColorSpace when loading images this way. We may be able to solve them in the future, but this PR still at least paves the way.

Sources/SnapshotTesting/Strategy/NSView.swift
```diff
@@ -0,0 +1,43 @@
+#if os(macOS)
+import Cocoa
+
+extension Strategy where Snapshottable == NSView, Format == NSImage {
```
stephencelis: I've reorganized a bit with `NSView` and `UIView`-specific files, while `View` contains some shared view code.

Sources/SnapshotTesting/Strategy/UIView.swift
```diff
@@ -0,0 +1,43 @@
...
+
+  public static func image(
+    precision: Float = 1,
+    size: CGSize? = nil,
+    traits: UITraitCollection = .init()
+    )
+    -> Strategy {
+
+      return SimpleStrategy.image(precision: precision).asyncPullback { view in
+        snapshotView(
```
stephencelis: This helper gets used for `UIView` and `UIViewController` to properly handle trait collection override, child view controller logic.

stephencelis:
Just to resummarize this PR:

1. Adds support for the safe area, and alongside it support for the iPhoneX, XR, XS Max.
2. Passes trait collection to layer renderer, allowing for people to render things at different scales/gamuts.
3. General refactor for less duplication between UIView and UIViewController strategies.

stephencelis:
> That would be awesome if possible. Currently we're having to force the use of the iPhone SE (or other 2x scale device) for UI snapshot testing (FBSnapshotTestCase has the same issue btw).

@mluisbrown I don't think this is possible, though our library will at least let you do something like this:

``` swift
assertSnapshot(matching: viewController, on: .iPhoneSe)
assertSnapshot(matching: viewController, on: .iPhone8)
assertSnapshot(matching: viewController, on: .iPhone8Plus)
assertSnapshot(matching: viewController, on: .iPhoneX)
assertSnapshot(matching: viewController, on: .iPhoneXr)
assertSnapshot(matching: viewController, on: .iPhoneXsMax)
assertSnapshot(matching: viewController, on: .iPadMini)
assertSnapshot(matching: viewController, on: .iPadPro10_5)
assertSnapshot(matching: viewController, on: .iPadPro12_9)
```

It'll generate trait-accurate screen shots for each device, but you'll need to run everything consistently from either a 2x simulator or a 3x simulator. Trying to use the 3x scale renderer from a 2x simulator leads to UIKit laying things out inaccurately, and in general things antialias a bit differently.

Issue #106, opened 2018-11-27
How to conform to DefaultSnapshottable


loudmouth:
Hey there,

First off, this library looks awesome and I really enjoyed reading the article you posted about it.

I was starting to integrate this into a project of mine and was intrigued by applying snapshots to instance of arbitrary types—this seems to be one advantage of this lib over FBSnapshotTestCase. I quickly found out that my types would need to conform to `DefaultSnapshottable` in order to assert snapshot matches for them. I'm fine implementing conformance via extensions in my test target, but I've yet to figure out the intent of the associated types and how I might implement conformance.

Let's say I wanted to test `NSAttributedString`, since in my case my library is rendering an NSAttributedString with a _lot_ of custom attributes. Any advice on how I would declare conformance to `DefaultSnapshottable` for `NSAttributedString`? Specifically, not sure what the associatedtype `Format` should be

loudmouth:
After doing some digging, I realize that implementing conformance might be tricky, and not something I have time for at the moment. I think conformance for `NSAttributedString` would be sufficient for my own use-case and if #52 is to be seen as a feature roadmap, it looks like it will be done in the future. If #52 is the feature roadmap, feel free to close this issue

mbrandonw:
> After doing some digging, I realize that implementing conformance might be tricky, and not something I have time for at the moment. I think conformance for `NSAttributedString` would be sufficient for my own use-case and if #52 is to be seen as a feature roadmap, it looks like it will be done in the future. If #52 is the feature roadmap, feel free to close this issue

Yeah attributed strings are definitely on our roadmap. The code I posted above is essentially what we'll be adding to the library, but right now we are heads down for the next week to get the 1.0 release of this library ready. We will get to it soon after that, or if you wanted to give it a shot that'd be great too!

loudmouth:
Wowzas, thanks for the ultra-comprehensive response.

Since I'm rendering to an `NSAttributedString` with custom attributes, i.e. attributes that are not native, but implemented by me, it turns out that using the `.dump` strategy is perfect. It gives me the ability to snapshot all the attributes at each respective range of characters.

I'm going to close this for now, as this issue was more of a forum-style question/answer.

Keep up the good work with this lib and thanks for the help!

mbrandonw:
@loudmouth that's great to hear! we would love to see how you end up using this if it's open source.

loudmouth:
Well it is open-source ;-)

[...]

PR #107, opened 2018-11-27
Use minus sign instead of hyphen minus

mbrandonw:
Currently if you copy and paste the diff from a failure message into something that can syntax highlight diffs, the deletions will not render. For example:

```diff
−a
+b
```

This is because we are using a hyphen minus instead of a minus sign. If we change to a regular minus sign it works:

```diff
-a
+b
```

### DIFF
--- a/Sources/SnapshotTesting/Diff.swift
+++ b/Sources/SnapshotTesting/Diff.swift
@@ -42,7 +42,7 @@ func diff<A: Hashable>(_ fst: [A], _ snd: [A]) -> [Difference<A>] {
   }
 }
 
-let minus = "−"
+let minus = "-"
 let plus = "+"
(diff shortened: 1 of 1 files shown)


stephencelis:
I wish it would render nicely in Xcode's font :(

Other than for the docs you were writing can you think of a reason to favor copy-pasting over reading?

mbrandonw:
I first encountered this while trying to write the blog post for next week and couldn't figure out what was happening. but then I came across in a real life situation in which I wanted to copy and paste a diff to #106, and then I figured out what was wrong.

I'm leaning more towards not merging this now, but it's a real bummer.

mbrandonw:
closing this for now, can revisit if it becomes a big problem.

PR #109, opened 2018-11-27
Decouple assertions from error messages

stephencelis:
This is needed to, for example, build Nimble matchers.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -37,98 +37,121 @@ public func assertSnapshot<A, B>(
   line: UInt = #line
   ) {
 
-  let recording = recording || record
+  let failure = verifySnapshot(
+    matching: value,
+    as: strategy,
+    named: name,
+    record: recording,
+    timeout: timeout,
+    file: file,
+    testName: testName,
+    line: line
+  )
+  guard let message = failure else { return }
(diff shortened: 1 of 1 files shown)


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -37,98 +37,121 @@ public func assertSnapshot<A, B>(
...
+        fileURLWithPath: ProcessInfo.processInfo.environment["SNAPSHOT_ARTIFACTS"] ?? NSTemporaryDirectory()
+      )
+      let artifactsSubUrl = artifactsUrl.appendingPathComponent(fileName)
+      try fileManager.createDirectory(at: artifactsSubUrl, withIntermediateDirectories: true)
+      let failedSnapshotFileUrl = artifactsSubUrl.appendingPathComponent(snapshotFileUrl.lastPathComponent)
+      try strategy.diffable.to(diffable).write(to: failedSnapshotFileUrl)
+
+      if !attachments.isEmpty {
+        #if !os(Linux)
+        XCTContext.runActivity(named: "Attached Failure Diff") { activity in
```
stephencelis: Still coupled to XCTest via XCTContext and XCTWaiter, but basically all testing in Swift piggybacks off XCTest these days anyway. We can revisit decoupling entirely if/when other test libraries become popular.

PR #111, opened 2018-11-29
Documentation

stephencelis:
Starting to add inline docs...


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -1,11 +1,27 @@
...
+///   - strategy: A strategy for serializing, deserializing, and comparing values.
+///   - name: An optional description of the snapshot.
+///   - recording: Whether or not to record a new reference.
+///   - timeout: The amount of time a snapshot must be generated in.
+///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
+///   - testName: The name of the test in which failure occurred. Defaults to the function name of the test case in which this function was called.
+///   - line: The line number on which failure occurred. Defaults to the line number on which this function was called.
 public func assertSnapshot<A, B>(
-  matching value: A,
+  matching value: @autoclosure () throws -> A,
```
stephencelis: Realized we can make this interface a lil nicer.

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -37,15 +53,16 @@ public func assertSnapshot<A, B>(
       identifier = String(counter)
     }
 
+    let testName = sanitizePathComponent(testName)
     let snapshotFileUrl = snapshotDirectoryUrl
-      .appendingPathComponent("\(testName.dropLast(2)).\(identifier)")
```
stephencelis: This assumed the implicit `#function` format like `testName()`. Added a proper slugging function instead.

Sources/SnapshotTesting/Diffable.swift
```diff
@@ -1,14 +1,30 @@
 import Foundation
 
-public struct Diffable<A> {
-  public let to: (A) -> Data
-  public let fro: (Data) -> A
-  public let diff: (A, A) -> (String, [Attachment])?
+/// The ability to compare `Value`s and convert them to and from `Data`.
+public struct Diffable<Value> {
```
stephencelis: We named the generics on `Strategy` but failed to do so here.

Sources/SnapshotTesting/Diffable.swift
```diff
@@ -1,14 +1,30 @@
...
+  /// Converts a value _to_ data.
+  public let to: (Value) -> Data
 
+  /// Produces a value _ from_ data.
+  public let fro: (Data) -> Value
```
stephencelis: @mbrandonw Think we should rename any of these? I took `to`/`fro` from a PureScript library because I thought it was cute, but we could expand to `from` or something even more verbose.

mbrandonw: yeah maybe just `toData` and `fromData`?

`to`/`fro` might be better names if we were to use an `Iso<A, B>` type.

Sources/SnapshotTesting/Strategy.swift
```diff
@@ -1,32 +1,55 @@
...
+/// A type representing the ability to transform a value into a diffable format (like text or an image) for snapshot testing.
 public struct Strategy<Snapshottable, Format> {
+  /// The path extension applied to references saved to disk.
   public var pathExtension: String?
+
+  /// How the snapshot format is diffed and converted to and from data.
   public let diffable: Diffable<Format>
+
+  /// How a snapshot is transformed to a diffable format.
   public let snapshotToDiffable: (Snapshottable) -> Async<Format>
```
mbrandonw: could this be renamed to just `snapshot`? it's the action of turn a snapshottable value into a diffable format.

I'm not sure "How a snapshot is transformed" reads well since it's not really the snapshot but rather the thing being snapshot.

Sources/SnapshotTesting/Strategy/UIView.swift
```diff
@@ -2,10 +2,18 @@
...
 
+  /// A snapshot strategy for comparing views based on pixel equality.
+  ///
+  /// - Parameters:
+  ///   - drawHierarchyInKeyWindow: Utilize the simulator's key window in order to properly render `UIAppearance` and `UIVisualEffect`s.
+  ///   - precision: The percentage of pixels that must match.
+  ///   - size: A view size override.
+  ///   - traits: A trait collection override.
   public static func image(
     drawHierarchyInKeyWindow: Bool = false,
```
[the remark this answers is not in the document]
stephencelis: I had a comment but it got verbose. Lemme give it another go.

Sources/SnapshotTesting/Strategy.swift
```diff
@@ -1,59 +1,90 @@
...
+    snapshot: @escaping (_ value: Snapshottable) -> Format
     ) {
     self.init(pathExtension: pathExtension, diffable: diffable) {
-      Async(value: snapshotToDiffable($0))
+      Async(value: snapshot($0))
     }
   }
 
-  public func asyncPullback<A>(_ transform: @escaping (A) -> Async<Snapshottable>) -> Strategy<A, Format> {
+  /// Transforms a strategy on `Value`s into a strategy on `A`s through a function `(A) -> Async<Value>`.
```
mbrandonw: It's more like:

```suggestion
  /// Transforms a strategy on `Snapshottable`s into a strategy on `A`s through a function `(A) -> Async< Snapshottable>`.
```

right? do you think people will be confused with the differing type names?

stephencelis: Ah yeah. Should we rethink the names?

mbrandonw: I guess `Strategy<Value, Format>` still reads really nicely.

Can't remember if we had a bigger reason for going with the super verbose `Snapshottable`. it's also not great that it ends in `-able`, makes it sound like a protocol.

Sources/SnapshotTesting/Strategy/UIView.swift
```diff
@@ -10,7 +10,7 @@ extension Strategy where Snapshottable == UIView, Format == UIImage {
   /// A snapshot strategy for comparing views based on pixel equality.
   ///
   /// - Parameters:
-  ///   - drawHierarchyInKeyWindow: Utilize the simulator's key window in order to properly render `UIAppearance` and `UIVisualEffect`s.
+  ///   - drawHierarchyInKeyWindow: Utilize the simulator's key window in order to render `UIAppearance` and `UIVisualEffect`s. This option requires a host application for your tests and will _not_ work for framework test targets.
```
mbrandonw: 👍

Sources/SnapshotTesting/Strategy.swift
stephencelis: I like `Value`! Can we also fix the naming on `Diffable`?

mbrandonw: `Diffing` would prob do. `Diffing<String>` and `Diffing<UIImage>` reads well.

mbrandonw: even `Snapshotting` is kinda nice instead of `Strategy` lol. but don't know if we want to change that.

PR #113, opened 2018-11-30
Fix unintentional truncation of custom testName parameter

### LINKED ISSUE #112: Last two characters of custom testName are being removed
Removing them makes sense if you are using `#function`, but when supplying a custom `testName` I don't think it makes quite as much sense.

Killectro:
Also removes the `line` parameter from the `verifySnapshot` method since it wasn't being used.

Fixes #112.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -22,8 +22,7 @@ public func assertSnapshot<A, B>(
     record: recording,
     timeout: timeout,
     file: file,
-    testName: testName,
-    line: line
+    testName: testName
   )
   guard let message = failure else { return }
   XCTFail(message, file: file, line: line)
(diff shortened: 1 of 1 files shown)


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -50,6 +48,7 @@ public func verifySnapshot<A, B>(
       let snapshotDirectoryUrl: URL = directoryUrl
         .appendingPathComponent("__Snapshots__")
         .appendingPathComponent(fileName)
+      let testName = testName.hasSuffix("()") ? String(testName.dropLast(2)) : testName
```
Killectro: The ideal solution here might be to do some sort of regex and rip out the entire thing, but given that test functions normally take no parameters I figured this would be a rare case.

stephencelis:
Ah yup! I noticed this yesterday 😬

Sorry about that! I've bundled a fix on this branch that does more general file name sanitization: https://github.com/pointfreeco/swift-snapshot-testing/pull/111/files#diff-4940eb0d5cb63d576fddd4e080b90475R168

I should have done a separate PR and shepherded it in sooner, though!

Killectro:
Thanks for the quick response as well!

PR #114, opened 2018-11-30
Add raw URLRequest strategy

stephencelis:


Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+extension Snapshotting where Value == URLRequest, Format == String {
+  public static let raw = SimplySnapshotting.lines.pullback { (request: URLRequest) in
```
stephencelis: Open to other naming, but this seemed appropriate.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -0,0 +1,18 @@
...
+
+    let status = "\(request.httpMethod ?? "GET") \(request.url?.absoluteString ?? "(null)")"
```
stephencelis: Should we render "(null)" or fatal error on an invalid request?

PR #117, opened 2018-12-03
More Documentation

stephencelis:


stephencelis:
Gonna merge for now. Can tweak as we go.

Documentation/Defining-Custom-Snapshot-Strategies.md
```diff
@@ -0,0 +1,82 @@
...
+      }
+  }
+}
+```
+
+#### Async Initializer
+
+`Snapshotting` defines an alternate initializer to describe snapshotting values in an asynchronous fashion.
+
+For example, were we to define a strategy for `WKWebView` _without_ [`asyncPullback`](#asyncpullback):
```
mbrandonw: does this need to be finished?

stephencelis: it could be fleshed out but it's fully descriptive of how one would define it.

mbrandonw: I wasn't sure what it was saying. I thought it was incomplete since that version of snapshotting web views won't work right?

stephencelis: Yeah though we support it automatically so I didn't think anyone would try to follow along and define it themselves. Is there another example that we could make up?

PR #119, opened 2018-12-03
Fix failing tests on alternative build platforms via snapshot variants

finestructure:
I've got a case in my own tests where there are small differences in rendering between my development machine and our CI machine. I suspect it's down to running macOS 10.14 on one and macOS 10.13 on the other.

The difficulty is that declaring either the "source of truth" will cause difficulties when working with the other. Obviously I don't want to declare the dev version as the "source of truth" as I'll never get the CI to pass. On the other hand if I take the CI variant I'll never be able to "pre-flight" test changes locally.

What I've therefore done is extend the matching to test a range of snapshot variants that are acceptable. It's perhaps slightly hacky but the way this works is that for a file `testSKView.macos.png` you can add variants `testSKView.macos.a.png` ... `testSKView.macos.z.png` which will also be tested (if the files exist).

It turns out that running the included macOS tests on my machine (10.14) also exhibit these deltas and I'm including my snapshots as variants.

There are the three diffs that failed for me previously but are now passing thanks to the variants:

screenshot 2018-12-03 at 09 17 03
screenshot 2018-12-03 at 09 15 29
screenshot 2018-12-03 at 09 11 44

These diffs looks quite significant but if you eyeball the images you'd be hard pressed to see a difference.

I saw that you recently also observed an image delta in a PR and I suspect that's same effect that I'm seeing between OS revisions.


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -123,11 +121,25 @@ public func verifySnapshot<A, B>(
...
-      let reference = snapshotting.diffing.fromData(data)
+      var allVariantsToTest = [snapshotFileUrl]
+      for c in "abcdefghijklmnopqrstuvwxyz" {
+        let url = snapshotUrl(directoryUrl: snapshotDirectoryUrl, testName: testName, identifier: identifier, pathExtension: snapshotting.pathExtension, variant: String(c))
+        if fileManager.fileExists(atPath: url.path) {
+          allVariantsToTest.append(url)
+        } else {
+          break
+        }
+      }
```
finestructure: It wasn't immediately obvious to me if/how this could be written with a `map` without testing for all files, i.e. I don't know if there's a way to break out of the map once the first file url does not exist.

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -123,11 +121,25 @@ public func verifySnapshot<A, B>(
...
+      for variantUrl in allVariantsToTest {
+        let data = try Data(contentsOf: variantUrl)
+        let reference = snapshotting.diffing.fromData(data)
 
-      guard let (failure, attachments) = snapshotting.diffing.diff(reference, diffable) else {
-        return nil
+        guard let (f, a) = snapshotting.diffing.diff(reference, diffable) else {
+          return nil
+        }
+        (failure, attachments) = (f, a)
```
finestructure: Likewise, this block feels a bit clumsy but if I try and map here it's rather hard or convoluted to break out of it.

finestructure:
If you find this a useful addition I'll add a note about the variants to the Readme.

There should probably also be a mention of variants (if there are any) in case of failures. I.e. in the presence of variants there should perhaps be a list of files that have been tested and that all have failed printed in the logs.

stephencelis:
@finestructure Thanks for the PR and self-review! I think having a better solution around this would be nice. We currently `#if`-disable view snapshots on certain CI boxes while we wait for them to update the OS. We don't have easy access to macOS 10.13, though, so generating variants for CI isn't necessarily a straightforward task for us.

Thinking over the problem, a coupla thoughts spring to mind:

- This kind of variance feels a bit view-strategy specific, so I wonder if strategies themselves should have some ability to adjust the snapshot name.

- I then wonder if it's possible to automatically and deterministically name variants by encoding the device scale, OS version, etc., instead. That way performance would be consistent and it should be less likely for stale variants to pile up over time. Currently you could encode variants manually using the `named` parameter, but an at-hand, deterministic way of doing so would be nice. Facebook's library has a `deviceAgnosticOptions` setting as a shortcut for this kind of thing.

@mbrandonw Any initial thoughts?

finestructure:
Hi Stephen, thanks for the quick response! Just a couple of thoughts triggered by it (sorry, it ended up getting lengthy...):

One thing to bear in mind is that I'm not 100% sure it's an OS variant that's causing the discrepancy. If it is, automating the capture by version as you outline would be nice.

However, I suspect there's probably always going to be some need to manually add other reference variants. I'm only starting to instrument our infrastructure and already I've encountered a few cases where I've triggered false positives due to domain intrinsic deltas (I'm currently snapshotting html pages):

- cookie or app banners
- cache invalidation name mangling (this affects `.lines` snapshots) (our js, svg, ... includes come with a random string to make sure they override caches on deployment)

In both cases I managed to work around it without variants by either injecting the appropriate cookies to making sure I get the expected result by demangling the html I'm fetching.

But it's probably inevitable that there'll be the need to provide the matcher with a notion of "fuzziness" one way or another. Arguably, variants are going to fall short here as well at some point (as they would in case of the cache name mangling, for instance).

Maybe there's a more general solution to be aimed for instead, like for instance excluding certain regions from a diff (regions being a `CGRect` in case of `.image` perhaps and a regex in case of `.lines` for example?).

On another note, you mention that you `#if`-disable CI tests because you cannot create the proper reference.

What I've done for our GitLab build is set the failures to be saved to a reachable directory which is archived with the build:

```
variables:
  SNAPSHOT_ARTIFACTS: $CI_PROJECT_DIR/Tests/WebpageTesterTests/__Snapshots__/failed
```

If I see that the variant that I pushed from my dev box causes a failure in the CI I go and download the failed files, compare them manually and if they're ok add them as variants to the build.

Maybe that's something that's helpful with Travis and CircleCI as well.

mbrandonw:
Hey @finestructure thanks for exploring this area, but yeah I think we can close this for now and reference it later if we need. I do think there is room for improvement for us to better support multiple OS's, but I believe we need to first make our `assertSnapshot` helper more extensible so that we don't keep cramming ad hoc logic into it. 

Thanks!

Issue #129, opened 2018-12-10
Add support for running on CI when “build for testing” agent is different to testing agent


rjchatfield:
Firstly, really enjoying Snapshot tests. Already found a bug in our JSON parsing logic. 

# Problem

We compile our tests on one agent, then run the tests on various other agents in parallel. At compile time, `#file` is an absolute path to the file. At run time (on a different agent with a different project path), that path is not valid.  Thus, all the tests try re-recording the snapshot and fail. 

# Suggestions

1. Allow to look in Bundle resources for the snapshot file
2. Allow a run-time file search to find the path of the snapshot

We’ve used resources for bundling JSON files in our test cases, so that is my preference here. 
Either option would be “opt-in”, defaulting to your folder structure based approach.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

vox-humana:
It could be environment variables like `IMAGE_DIFF_DIR` and `FB_REFERENCE_IMAGE_DIR` in [iOSSnapshotTestCase](https://github.com/uber/ios-snapshot-test-case/).

vox-humana:
And just one small related question. Why there is `StaticString` for `file` parameter in [`assertSnapshot`](https://github.com/pointfreeco/swift-snapshot-testing/blob/27315eb6e29dbb346840b9653112976921abd389/Sources/SnapshotTesting/SnapshotTestCase.swift#L31) instead of `String`? So there is no way to pass a run-time requested string even from the caller's side.

stephencelis:
Thanks for reporting the issue (and PR, which I'll look at shortly)! Re: your suggestions:

> 1. Allow to look in Bundle resources for the snapshot file
> 2. Allow a run-time file search to find the path of the snapshot

Either of these sound like viable solutions, though we wanna think over the ramifications a bit.

> Why there is `StaticString` for `file` parameter in [`assertSnapshot`](https://github.com/pointfreeco/swift-snapshot-testing/blob/27315eb6e29dbb346840b9653112976921abd389/Sources/SnapshotTesting/SnapshotTestCase.swift#L31) instead of `String`?

I think this is an oversight that we can fix.

stephencelis:
@vox-humana Unfortunately `file` needs to be passed to `XCTFail` as a `StaticString`, so this interface can't be changed.

vox-humana:
@stephencelis Oh, I see, it is sad. Thank you.

ccarpita:
[...]

PR #133, opened 2018-12-11
Add assertSnapshotFromResources() to allow running on CI

rjchatfield:
This is a crude duplication of the assertSnapshot() API to solve this issue I raised yesterday: #129 

We're keen to use SnapshotTesting to our repo, but using an absolute filepath from `#file` won't work for us.


stephencelis:
Thanks for the PR! I think ideally we can find the right abstraction for `verifySnapshot` to avoid the duplication. Lemme chat with @mbrandonw about potential solutions here.

rjchatfield:
That's fine. I wanted to try pass in a "verify snapshot strategy" but I couldn't make anything ergonomic enough that I liked. So we've implemented this PR locally in our codebase. The only thing we can't implement is storing the attachments in the XCTActivity.

```swift
XCTContext.runActivity(named: "Attached Failure Diff") { activity in
    attachments.forEach {
        activity.add($0.rawValue) // `rawValue` is internal
    }
}
```

Any chance you could make `Attachment.rawValue` public?

rjchatfield:
Yeah that looks good. Thanks.

ecunha-ta:
This matter is mentioned in another [issue](https://github.com/pointfreeco/swift-snapshot-testing/issues/129) opened on pointfree repo. 

Any news if something like is suggested on this MR is going to be merged?

PR #134, opened 2018-12-11
Assert all snapshot have been checked

szotyi:
This is a feature request for testing, that all snapshots have been checked for a given test case. We have a large code base with dynamic snapshot tests and frequent refactors. This helps us better maintain our snapshots and find more mistakes. It is only for mac/iOS currently.

I hope you find it usefull (it would be nice, not to maintain a fork to use your project 🙂) and we can start a discussion on it.

Example usage: 
```swift
class MyViewControllerTests: XCTestCase {
  override func tearDown() {
    assertAllSnapshotsChecked(for: MyViewControllerTests.self)
    super.tearDown()
  }
  // ...
}
```

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -195,10 +195,16 @@ public func assertAllSnapshotsChecked(for testClass: XCTestCase.Type, file: Stat
 
   let fileManager = FileManager.default
   do {
-    let expected = try fileManager.contentsOfDirectory(
-        at: snapshotDirectoryUrl, includingPropertiesForKeys: [], options: .skipsHiddenFiles)
-    let diff = expected.filter { !checked[snapshotDirectoryUrl, default: []].contains($0) }
-    XCTAssertEqual(diff, [], "Not all files were checked", file: file, line: line)
+    let expected = try fileManager.contentsOfDirectory(at: snapshotDirectoryUrl, 
+                                                       includingPropertiesForKeys: [], 
+                                                       options: .skipsHiddenFiles)
+    let diff: String = expected
+                         .filter { !checked[snapshotDirectoryUrl, default: []].contains($0) }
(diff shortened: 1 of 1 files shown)


szotyi:
I gave it a shot for the "platform as folder", you can check it out at: https://github.com/szotyi/swift-snapshot-testing/commit/6e1bea20553d9500b78e633613c0a4fc330be646

(My questions from the previous comment are still up)

stephencelis:
> Good catch. Platforms make this difficult. To annotate snapshots in meaningful and storable manner, I can only see creating folders for each platform or using naming schemes (eg. postfix). This would create duplicated snapshots and would have to create some kind of migration functionality from old versions of the library. For the first problem maybe some kind of way to create "combined platforms" or if there is only one than no annotation (the current way).
> Do you have other ideas?

A global configuration feature around annotation would be helpful in general, though I'm not sure what it would look like quite yet.

> What other edge cases do you see (beside the Linux compatibility)?

None that come to mind right now, but I'll think on it some more.

> What’s your strategy of new features? Do you only put fully implemented ones in or can there be ones with limited functionality (with clear documentation) and than approve them over time (when demand arrives or fresh ideas pop up)?

Nothing formalized yet, but I think the thoughts are:

- New features should be non-breaking (till a new major release, e.g., 2.0)
- Additive new features should be rolled out in minor releases (e.g., 1.1) and should be stable/documented by the time of the release
- New features are welcome into master without having been fully stabilized/documented for those that want to test the feature out

We'd also like to keep the core library small, I think, but extensible. Ideally your helper could be written as another library and somehow hook into our public API. Then over time, popular external libraries could be folded into this main one.

szotyi:
Was thinking about your “requirements” and couldn’t come up with a nice way, that would fit the robustness and easiness of your current state. 
I will keep eye on your future enhancements and features (and of course will watch you videos) and if any new idea comes to mind will let you know. 
Feel free to close this pr (I will maintain my fork). 
Merry Christmas!

PR #143, opened 2018-12-15
Avoid "swift test" crash when a snapshot doesn't match

### LINKED ISSUE #121: Activities are not available in SwiftPM projects (even on macOS) when using `swift test`
Hey folks, ran out of my timebox for this one, but I'm running on macOS, but using SwiftPM. This means one of the `XCTContext.runActivity` is running, but the closed-source XCTest isn't available. This means `#if !os(Linux)` isn't going to work when running via `swift test` . 

[Their check](https://github.com/apple/swift/blob/master/stdlib/public/SDK/XCTest/XCTest.swift#L35) probably isn't available for us to use, if you want to test it out:

```sh
git clone https://github.com/danger/danger-swift.git
cd danger-swift
git checkout snapshots # or dc41e4e06eb2c994e0ece6fdc355386a212fc0fd
swift build
swift run komondor install
swift package generate-xcodeproj
open Danger.xcodeproj

swift test
```

stephencelis:
Fixes #121.

We haven't hit this because we haven't ever troubleshot a failing snapshot while running `swift test`. I don't love the solution but not sure we can do better without XCTest improving their API:

https://bugs.swift.org/browse/SR-9526


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -154,9 +154,11 @@ public func verifySnapshot<Value, Format>(
 
       if !attachments.isEmpty {
         #if !os(Linux)
-        XCTContext.runActivity(named: "Attached Failure Diff") { activity in
-          attachments.forEach {
-            activity.add($0)
+        if ProcessInfo.processInfo.environment.keys.contains("__XCODE_BUILT_PRODUCTS_DIR_PATHS") {
```
stephencelis: Relying on a private Xcode implementation detail :/

stephencelis: There are a few other keys we could rely on instead, like `XCTestConfigurationFilePath`...

PR #144, opened 2018-12-16
Use official Swift docker image

slashmo:
Since the docker images for Swift moved to Apple's GitHub organization the `Dockerfile` can be simplified to use the official image. (The [4.2 tag](https://github.com/apple/swift-docker/blob/master/4.2/Dockerfile) has Swift 4.2.1 installed).


stephencelis:
Thanks for the PR!

PR #145, opened 2018-12-17
Attributed Strings

stephencelis:
Don't need to merge this immediately. Was just having some fun.


Sources/SnapshotTesting/Snapshotting/NSAttributedString.swift
```diff
@@ -0,0 +1,37 @@
+#if canImport(Cocoa)
+import Cocoa
+
+extension Snapshotting where Value == NSAttributedString, Format == NSImage {
+  public static func image(precision: Float = 1, maxWidth: CGFloat) -> Snapshotting {
```
stephencelis: Could also do an `html` strategy...

Sources/SnapshotTesting/Snapshotting/NSAttributedString.swift
```diff
@@ -0,0 +1,37 @@
...
+      label.frame.size = label.fittingSize
+      return label
+    }
+  }
+}
+#elseif canImport(UIKit)
+import UIKit
+
+extension Snapshotting where Value == NSAttributedString, Format == UIImage {
+  public static func image(precision: Float = 1, maxWidth: CGFloat) -> Snapshotting {
```
stephencelis: Is `maxWidth` right?

mbrandonw: I think just `width` would be fine since the image probably will be that exact width.

stephencelis:
Closing this out to revisit some other time.

PR #150, opened 2018-12-19
Add support for multiple snapshotting strategy for each assert

### LINKED ISSUE #116: Group helper: assertSnapshots(matching:as:)
Would be nice to be able to assert on multiple snapshot strategies at once, as we do with a helper here here:

https://github.com/pointfreeco/pointfreeco/blob/8feb17812b2f076ab4996450ee85583ec77f6f54/Tests/PointFreeTests/AboutTests.swift#L20-L26

With it defined, we can also make an extension for some common things.

``` swift
extension Dictionary where Key == String, Value == Snapshotting<UIViewController, UIImage> {
  static let allDevices: Dictionary = [
    "iphone-se-portrait": .image(on: .iPhoneSe(.portrait)),
    // …
  ]
}

// …

assertSnapshots(matching: vc, as: .allDevices)
```

mackoj:
Attempt to fix #116


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -30,17 +30,49 @@ public func assertSnapshot<Value, Format>(
   testName: String = #function,
   line: UInt = #line
   ) {
+  assertSnapshot(matching: value,
+                 as: [snapshotting],
+                 named: name,
+                 record: recording,
+                 timeout: timeout)
```
mbrandonw: Thanks for the PR! Stephen and I will discuss this sometime soon, but in the meantime there is a quick thing you can do to get tests passing. Make sure to pass along the `file`, `testName` and `line` to the assert snapshot helpers so that files names are recorded properly.

We also prefer to have all the function arguments on a newline if they don't fit on online, as opposed to this hanging argument style.

Also there's an extra image that was added to the repo that could be deleted:

`assertSnapshot-matching-as-named-record-timeout-file-testName-line.1.png`

mackoj: Thanks for the comments, I will do the modification

stephencelis:
Hey @mackoj, thanks for the PR! One thing that I think needs to change is that `assertSnapshots` needs to lose the `named` parameter. The `name` there replaces the auto-incrementing default name of `1`, `2`, etc. With that in mind `assertSnapshots` will have to call `assertSnapshot` instead.

We wrote these helpers in the Point-Free repo awhile back:

https://github.com/pointfreeco/pointfreeco/blob/18ec4ac7486f9c817d31e20dbebf0cc597e6adba/Sources/PointFreeTestSupport/PointFreeTestSupport.swift#L558-L603

Feel free to grab em and use em in this PR! Will be happy to merge when you do 😄

And thanks for taking the time to help!

stephencelis:
Oh! And unfortunately we have to define these helpers twice right now (you can scroll above the last link to see them defined inside the Linux `SnapshotTestCase` class). Linux Foundation currently lacks the `XCTWaiter` APIs, so we need to use a subclass there.

mackoj:
Ok I will fix it today. Thanks for the comment @stephencelis

stephencelis:
Thanks again! 😄

PR #153, opened 2018-12-20
Move around some failure messaging.

mbrandonw:
Re-prioritizing some of the failure messaging so that the shorter stuff is on top and doesn't get knocked down by the diff.

### Failure

messages image 3326583461

### Record

messages image 3768075003


Sources/SnapshotTesting/Snapshotting/String.swift
```diff
@@ -21,6 +21,6 @@ extension Diffing where Value == String {
       .flatMap { [$0.patchMark] + $0.lines }
       .joined(separator: "\n")
     let attachment = XCTAttachment(data: Data(failure.utf8), uniformTypeIdentifier: "public.patch-file")
-    return ("Diff: …\n\n\(failure)", [attachment])
+    return (failure, [attachment])
```
mbrandonw: maybe this `"Diff: …"` isn't necessary anymore. the screenshots looked ok.

PR #154, opened 2018-12-20
Change pullback generic names

stephencelis:
The `A` can make the pullback concept a lil harder to understand, I think. How about we ground it in a more familiar concept by calling it `Root`? It's not quite the same as `Root`-`Value` of key paths, but it's a similar relationship.


mbrandonw:
I personally prefer the `A` because root isn't quite right. Given the `(NSAttributedString) -> UILabel` function we wouldn't think of `NSAttributedString` as a "root" of `UILabel`.

We can also just do `NewValue`.

PR #157, opened 2018-12-22
Proof-of-concept: pretty print URLRequest body when JSON

mbrandonw:
A few things to consider:

* Should this be baked into `.raw` or should it have another name so that people can choose. This might be important if people really want to snapshot exactly what is in the URLRequest.

* Should we check for `application/json` or should we always try pretty printing?


Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -3,16 +3,29 @@ import Foundation
...
     let headers = (request.allHTTPHeaderFields ?? [:])
       .map { key, value in "\(key): \(value)" }
       .sorted()
 
-    let body = request.httpBody
-      .map { ["\n\(String(decoding: $0, as: UTF8.self))"] }
-      ?? []
+    let body: [String]
+    do {
+      if #available(iOS 11.0, *) {
```
stephencelis: Aha, needs `macOS 10.3` and `tvOS 11.0` in here.

mbrandonw: Oof good catch. Will get to next time I have internet!

stephencelis:
> Should we check for `application/json` or should we always try pretty printing?

Maybe even more general testing a `"\bjson\b"` regex would work, since people may be using things like `text/json` or `application/json; charset=utf8`.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
stephencelis: I gotchu

Issue #151, opened 2018-12-19
The file "Test.swift" doesn’t exist


hoppus:
Can not make SnapshotTesting work.
When I use `assertSnapshot(matching: view, as: .image)` I get this message "The file "failed "Test.swift" doesn’t exist"

What I can do ?

stephencelis:
Hi @hoppus, thanks for the report! We haven't heard of anyone hitting this issue. Can you provide step-by-step instructions on how to reproduce the failure and/or attach a sample project exhibiting the issue?

hoppus:
```
SnapshotTestApp[31552:1158378] Failed to load test bundle from file:///Users/eugenypopov/Library/Developer/Xcode/DerivedData/SnapshotTestApp-ezokglikaewvzbdydofehynpzvvd/Build/Products/Debug-iphonesimulator/SnapshotTestApp.app/PlugIns/SnapshotTestAppTests.xctest/: Error Domain=NSCocoaErrorDomain Code=3587 "dlopen_preflight(/Users/eugenypopov/Library/Developer/Xcode/DerivedData/SnapshotTestApp-ezokglikaewvzbdydofehynpzvvd/Build/Products/Debug-iphonesimulator/SnapshotTestApp.app/PlugIns/SnapshotTestAppTests.xctest/SnapshotTestAppTests): Library not loaded: @rpath/libswiftGLKit.dylib
  Referenced from: /Users/eugenypopov/Library/Developer/Xcode/DerivedData/SnapshotTestApp-ezokglikaewvzbdydofehynpzvvd/Build/Products/Debug-iphonesimulator/SnapshotTestApp.app/PlugIns/SnapshotTestAppTests.xctest/Frameworks/SnapshotTesting.framework/SnapshotTesting
[...]

PR #158, opened 2018-12-22
Improve CocoaPods instructions

stephencelis:
I think it's good to show more Podfile structure given that we're a _test_ target dependency.


stephencelis:
(Hopefully this is also the fix for #151.)

PR #167, opened 2019-01-31
Add option to provide own reference directory for images

sunshinejr:
Hey guys! 

First of all thanks for this awesome library, had used Jest for a while and this was something I wished iOS world had as well.

I was migrating from some other snapshot library and we were missing one environment variable that we were using - custom directory for recorded snapshots.

I named it `SNAPSHOT_REFERENCE_DIR`, but let me know if you think there is a more suitable name for this one.

zrzut ekranu 2019-01-31 o 14 04 55

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -141,8 +141,10 @@ public func verifySnapshot<Value, Format>(
     do {
       let fileUrl = URL(fileURLWithPath: "\(file)")
       let fileName = fileUrl.deletingPathExtension().lastPathComponent
-      let directoryUrl = fileUrl.deletingLastPathComponent()
-      let snapshotDirectoryUrl: URL = directoryUrl
+      let userDefinedDirectory = ProcessInfo.processInfo.environment["SNAPSHOT_REFERENCE_DIR"]
+      let fallbackDirectoryUrl = fileUrl.deletingLastPathComponent()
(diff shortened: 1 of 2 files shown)


sunshinejr:
@stephencelis of course! Let me know once you decide on something 👍

bitomule:
Any update on this? I'm really interested.

mbrandonw:
Hi @sunshinejr, Stephen and I just discussed this and we think it's an important thing to customize. I just pushed this commit 6db576b that shows what we are thinking for it. 

The function that really powers all the snapshot work is `verifySnapshot`. The `assertSnapshot` functions just call it under the hood, and so it's `verifySnapshot` that we imagined 3rd parties could use to write their own assert helpers on top of.

So, we would allow `verifySnapshot` to be provided a custom `snapshotDirectory`, which it would use instead of building up `{fileDirectory}/__Snapshots/{fileName}`. Then you could write your own assert helpers that call `verifySnapshot` with any directory structure you want. It could come from the environment, and it could even be made to match nearly exactly what iOSSnapshotTestCase does.

Do you want to take over the work I started in that commit (you can cherry pick or copy-paste it), or would you like us to finish it off (docs and the linux version)?

sunshinejr:
@mbrandonw yeah that's a nice one 👍 I think you guys will do a better job in finishing this one so I'll leave it to you! Thanks!

mbrandonw:
@sunshinejr no prob, I just opened this pr: https://github.com/pointfreeco/swift-snapshot-testing/pull/170

PR #169, opened 2019-02-03
Create a typealias for SnapshotTestCase when not on Linux

f-meloni:
Hey 👋 
Thank you for this library, is really cool and easy to use :)
While i was implementing some tests on a project that is both `macOS` and `Linux`, I came to a point where i had to create a `typealias` in order to make the tests work on both the platforms without have to duplicate the code (and i saw you did the same on your tests).
This is why i though that adding a public `typealias` would probably solve the problem for everyone that is using the library :)

### DIFF
--- a/Sources/SnapshotTesting/SnapshotTestCase.swift
+++ b/Sources/SnapshotTesting/SnapshotTestCase.swift
@@ -1,7 +1,7 @@
 import XCTest
 
 #if !os(Linux)
-@available(swift, obsoleted: 5.0, message: "Please use XCTestCase instead")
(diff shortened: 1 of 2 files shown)


f-meloni:
@stephencelis thank you for the review :)
Yes you can deprecate a `typealias`, if it makes sense for you I could change it with something like

```
@available(swift, obsoleted: 5.0, message: "Please use XCTestCase instead")
public typealias SnapshotTestCase = XCTestCase
```

mbrandonw:
Yeah this makes sense to me!

f-meloni:
Nice :)
Updated!

PR #179, opened 2019-02-21
Fix syntax in strategy documentation

BasThomas:


mbrandonw:
Good catch, thanks @BasThomas!

Issue #182, opened 2019-03-01
[Feature] Store snapshots from multiple simulators side-by-side


tikitu:
We're just starting to use this wonderful library, and already we're scope-creeping it to solve problems it was never intended to...

Namely: it looks like it will do a beautiful job of cross-OS-version/cross-device regression testing, except for the fact that the snapshot filename doesn't depend on any of the simulator details. I see #128 is related, where you [suggest the workaround](https://github.com/pointfreeco/swift-snapshot-testing/issues/128#issuecomment-445230865) of naming the file by hand, but the problem is more general than just display scale: in particular we're interested in regression testing for OS behaviour changes over different versions, and the by-hand approach gets out of control pretty quickly.

I'm wondering if we could do something more systematic, by encoding some relevant details of the simulator setup in the filename: definitely OS version and device display scale, there might be more. I guess it should be opt-in, or use the current naming scheme as a fallback, or something similar.

I'd be happy to take a crack at a PR if y'all would be open to this idea, what do you think?

stephencelis:
Hey @tikitu!

Yeah this is coming up often enough that we definitely think it needs to be baked in. It'll be a breaking change so let's add a new global configuration to opt into it, and then we can enabled it by default in SnapshotTesting 2.0.

We'd want to bake in:

- OS
- Color gamut
- Device scale
- (Anything else?)

Some examples:

```
// unnamed first (index 1) snapshot of "testMyView" test, iOS 12.1, SRGB gamut
testMyView-1-iOS-12.1-srgb@2x.png 

// "customName"-named snapshot of "testMyView" test, iOS 12.0, P3 gamut
testMyView-customName-iOS-12.0-p3@2x.png
```

This means the current delimiter is changing from `.` to `-`, but it makes the OS version read a bit more nicely.

One other requirement: we should only record automatically if _no_ screen shot exists for this test. If you run on a simulator, record a snapshot, and then run on an incompatible simulator, it should _not_ record another snapshot, it _should_ fail with an error about an existing snapshot for a different simulator type. In order to recover the ability to record for multiple simulators, one could either:

1. Explicitly set `record = true`
2. We could introduce another global, `SnapshotTesting.supportedPlatforms`, that turns auto-record on for some explicit platforms.

Whatcha think?

tikitu:
[...]

PR #183, opened 2019-03-07
Filenames include simulator details

tikitu:
Provides #182 

Proof-of-concept for a heading check, this is far from complete.

Still to do:
* [ ] Everything except iOS-on-Mac is new to me: MacOS, tvOS, Linux, etc need attention
* [ ] All FIXMEs removed
* [ ] Moar tests
* [ ] Make the simulators used for CI explicit (because they matter for the filenames)
* [ ] Documentation
* [ ] Another PR for v2, because all the test-snapshot filenames will need updating...


tikitu:
Particular things I'd like to get feedback on:

1. I used the same mechanism for opt-in (pre-v2) and for auto-recording on a new simulator:  `SnapshotTesting.supportedPlatforms = []` turns on the new filenames and otherwise behaves like v1 (switching simulators will fail tests, although more explicitly), but if you add platform strings in `supportedPlatforms` you get auto-first-snapshot-creation, without `record=true`, for those platforms. This is probably clearer in [code](https://github.com/pointfreeco/swift-snapshot-testing/pull/183/files#diff-4940eb0d5cb63d576fddd4e080b90475R249) or [test](https://github.com/pointfreeco/swift-snapshot-testing/pull/183/files#diff-d1c0c8ce08d2a1f90785bcfc178d2ebaR602) form. Okay? Nokay?
2. (As I imagine it v2 removes optionality from `SnapshotTesting.supportedPlatforms` so opting in is no longer needed, but leaves the rest of the code alone. Ok? Better ideas?)
3. Scale and gamut are image-specific, while iOS-version in principle could affect anything but will most likely affect `UIKit` stuff I guess. I thought perhaps to ask `Snapshotting<Value, Format>` if it wants per-platform filenames, with some sensible defaults: what do you think?
4. Updating to v2 will be a right pain if you've got a lot of screenshots: shall we be a bit helpful? I thought perhaps to special-case the autorecord behaviour [if it finds a v1 filename](https://github.com/pointfreeco/swift-snapshot-testing/pull/183/files#diff-4940eb0d5cb63d576fddd4e080b90475R247): diff new file against old, only store if the diff matches, and give a more explicit autorecord-failure including the old filename for easy deletion.

stephencelis:
Hi @tikitu! Thanks for spending time on this! I haven't had time to dig in too deeply, but lemme see if I can answer your questions.

> I used the same mechanism for opt-in (pre-v2) and for auto-recording on a new simulator: SnapshotTesting.supportedPlatforms = [] turns on the new filenames and otherwise behaves like v1 (switching simulators will fail tests, although more explicitly), but if you add platform strings in supportedPlatforms you get auto-first-snapshot-creation, without record=true, for those platforms. This is probably clearer in code or test form. Okay? Nokay?

This use of `nil` makes sense to me. I wonder if we can make `supportedPlatforms` less stringly-typed, though? Maybe a `RawRespresentable` struct wrapping a string so we can ship with some statics?

> (As I imagine it v2 removes optionality from SnapshotTesting.supportedPlatforms so opting in is no longer needed, but leaves the rest of the code alone. Ok? Better ideas?)

Sounds good!

> Scale and gamut are image-specific, while iOS-version in principle could affect anything but will most likely affect UIKit stuff I guess. I thought perhaps to ask Snapshotting<Value, Format> if it wants per-platform filenames, with some sensible defaults: what do you think?

On the one hand, `pathExtension` could incorporate `Format`-specific things on `Snapshotting`, but on the other hand, scale and gamut _could_ affect snapshots of text that refer to the device scale and gamut...

> Updating to v2 will be a right pain if you've got a lot of screenshots: shall we be a bit helpful? I thought perhaps to special-case the autorecord behaviour if it finds a v1 filename: diff new file against old, only store if the diff matches, and give a more explicit autorecord-failure including the old filename for easy deletion.

Auto-migration would be nice but maybe not worth it. Deleting `__Snapshots__`, re-recording, and adding to Git should show all the moves.

tikitu:
> `Format`-specific things on `Snapshotting`

I had in mind to also allow per-assert overriding, but looking more closely the idea seems not to play at all nicely with the compositional structure of `Snapshotting` (e.g. a `pullback` might change the platform-sensitivity): let's drop that one.

I hadn't thought of deleting `__Snapshots__` for a migration, that's much simpler and needs no code support at all: sold!

I'll take a look at a `RawRepresentable` platform struct and do some tidying. Thanks for the feedback!

tikitu:
@stephencelis Turns out I'm not 100% sure I understand what you were imagining for removing the stringly-typing of supported platforms. Here's [an attempt](https://github.com/pointfreeco/swift-snapshot-testing/pull/183/commits/8850eb1e5cdb9f944d7faab44c6527dfd0cc5948), but it's pretty heavyweight: did you have something simpler in mind? I feel like I might be missing something here...

stephencelis:
> Turns out I'm not 100% sure I understand what you were imagining for removing the stringly-typing of supported platforms. Here's an attempt, but it's pretty heavyweight: did you have something simpler in mind? I feel like I might be missing something here...

I see what you mean, but it's not so bad. @mbrandonw and I need to take some time and walk through this PR (and some of the other issues reported here the past couple weeks). Will try to get back to you soon!

tikitu:
If that's on the right track, all good! I have to put my efforts here on pause for a week or so unfortunately, but I hope to be able to come back to it next week.

gabrieloc:
Any updates on this?

tikitu:
> Any updates on this?

Hooooo boy... I would love to come back and take another crack at this, but I honestly can't say when that will be possible.

stephencelis:
Ha if only there were all the time in the world! @tikitu How do you feel about the current status of this? Should we dive into it and check things out again or put this on hold, closing it till we have time to revisit?

tikitu:
@stephencelis If you could take a look that would be great: if you're comfortable with the direction it's going I can commit some time before the end of this month to bring it up to date and handle the "I know how it should go just didn't do it yet" marks.

(Given my previous track record, if we get to June and I didn't make any update please give it a dishonourable retirement.)

lottadot:
What about having `orientation` be in the snapshot filename? 
Additionally `locale` if your app has multiple languages?

PR #185, opened 2019-03-09
Proof of concept - Informing

i-dama:
👋  While working with SnapshotTesting and mostly comparing images doing iOS-on-mac, I realized that my workflow isn't where I would like it to be. I found myself moving the cursor around a lot to see the the attachements in Xcode, or selecting the commands to open a custom diff tool. I noticed it the most in 3 scenarios:

- 1. Running the test for the first time - The reference is automatically recorded 🔝 . I like to check the image myself before commiting my changes, so I copy-paste the path to see the image
- 2. Comparing a failing test - All images are attached in Xcode 💯 . Seeing the attachements requires some navigation around Xcode, or copying the produced command.
- 3. Recording a new reference - It's as simple as setting `record = true` and re-running the test. The "new" reference already exists after the first run in the form of the failing snapshot, so I thought to myself - why not just use it?

To address these, I started experimenting on a POC for a Mac app that I would run alongside Xcode. The app would "listen" for the failing tests, and it would display the reference and the failing snapshot in realtime. Changing the reference wouldn't require a record re-run. Here is a GIF of it in action:

Viewer POC

To get to the POC, I played around with a fork of SnapshotTesting in which I modified the parts that inform on reference recordings and test failures. I was wondering if such a similar change would be acceptable as part of the main repo.
In this PR, I added a concept of how that might look:

The `Informing` struct is the ability to inform the user (me, yay) about new reference recordings and failing tests. The functions return `String` so it's easy to plug in the existing code, and print them alongside the failure message.
It would "formalize" the `diffTool` a bit, and allow for more variety via plug-ins. Right now, we can only extend the `diffTool` by providing a string value that would get printed in the output. Plug-ins would be able to add custom code that would process new recordings or test failures.

The struct is a generic on `Format`, but it doesn't use the type infromation in it's functions. The generic restriction is there to allow plugins to be created on certain `Format`s only. For example, I might want to open Kaleidoscope on test failures when the format is `Image`. That plug-in would be defined on the `Image` format.

The default `Informing` uses the existing behavior - it prints out the paths to the files.
As an example, I added `ksdiff` as a plug-in, but it would need to be reverted to `diffTool` to be backwards compatible.

What do you think?

Thanks for the great library! 😍  SnapshotTesting and Witness Oriented Design!

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -1,11 +1,6 @@
 #if !os(Linux)
 import XCTest
 
-/// Enhances failure messages with a command line diff tool expression that can be copied and pasted into a terminal.
-///
-///     diffTool = "ksdiff"
-public var diffTool: String? = nil
-
 /// Whether or not to record all new references.
 public var record = false
 
@@ -23,6 +18,7 @@ public var record = false
 public func assertSnapshot<Value, Format>(
   matching value: @autoclosure () throws -> Value,
   as snapshotting: Snapshotting<Value, Format>,
+  informing: Informing<Format> = .basic,
   named name: String? = nil,
   record recording: Bool = false,
   timeout: TimeInterval = 5,
@@ -34,6 +30,7 @@ public func assertSnapshot<Value, Format>(
   let failure = verifySnapshot(
     matching: value,
     as: snapshotting,
+    informing: informing,
     named: name,
     record: recording,
     timeout: timeout,
@@ -58,6 +55,7 @@ public func assertSnapshot<Value, Format>(
 public func assertSnapshots<Value, Format>(
   matching value: @autoclosure () throws -> Value,
   as strategies: [String: Snapshotting<Value, Format>],
+  informing: Informing<Format> = .basic,
   record recording: Bool = false,
   timeout: TimeInterval = 5,
   file: StaticString = #file,
@@ -69,6 +67,7 @@ public func assertSnapshots<Value, Format>(
     assertSnapshot(
       matching: value,
       as: strategy,
+      informing: informing,
       named: name,
       record: recording,
       timeout: timeout,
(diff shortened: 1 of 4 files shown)


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -1,11 +1,6 @@
 #if !os(Linux)
 import XCTest
 
-/// Enhances failure messages with a command line diff tool expression that can be copied and pasted into a terminal.
-///
-///     diffTool = "ksdiff"
-public var diffTool: String? = nil
```
stephencelis: Any idea on how to restore this bit of global configuration? The `Format` generic makes it tricky, but maybe we could use some kind of lookup table...

``` swift
SnapshotTesting.inform[Data.self] = .basic
SnapshotTesting.inform[String.self] = .basic
SnapshotTesting.inform[UIImage.self] = .basic
```

Alternately we could drop the generic for now till we have a compelling reason to introduce it...

i-dama: I think we should not delete the `diffTool` var and keep backwards compatibility.

The `.basic` Informing would do the same as the current behavior:
```
diffTool
	        .map { "\($0) \"\(snapshotFileUrl.path)\" \"\(failedSnapshotFileUrl.path)\"" }
	        ?? "@\(minus)\n\"\(snapshotFileUrl.path)\"\n@\(plus)\n\"\(failedSnapshotFileUrl.path)\""
```
Informing would default to `.basic` and the behavior would be unchanged.


Regarding the global configuration, a lookup table should work. I might be focusing too much on the use case I have I mind, but it might be useful to be able to configure different diff tools for different Formats.

Let me know when you figure out the next steps 👍

i-dama: @stephencelis Any update? :D

PR #195, opened 2019-03-23
Faster sort function

rjchatfield:
Found a small change of building an array of snaps, *then* sorting them, it reduced on of our test cases from 6m 8s to just 9s. And noticed it crept over 2GB of memory usage.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/Any.swift
+++ b/Sources/SnapshotTesting/Snapshotting/Any.swift
@@ -55,7 +55,12 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
 }
 
 private func sort(_ children: Mirror.Children) -> Mirror.Children {
-  return .init(children.sorted { snap($0) < snap($1) })
+  return .init(
+    children
+      .map({ (child: $0, snap: snap($0)) })
+      .sorted(by: { $0.snap < $1.snap })
+      .map({ $0.child })
+  )
 }
 
 /// A type with a customized snapshot dump representation.


rjchatfield:
image

Leaving that here for reference.

Sources/SnapshotTesting/Snapshotting/Any.swift
```diff
@@ -55,7 +55,12 @@ private func snap<T>(_ value: T, name: String? = nil, indent: Int = 0) -> String
 }
 
 private func sort(_ children: Mirror.Children) -> Mirror.Children {
-  return .init(children.sorted { snap($0) < snap($1) })
+  return .init(
+    children
+      .map({ (child: $0, snap: snap($0)) })
+      .sorted(by: { $0.snap < $1.snap })
+      .map({ $0.child })
```
stephencelis: I wonder if we can simplify further if we:

1. Remove this `sort` function entirely
2. Change `var children = mirror.children` to `var children = mirror.children.map { snap($1, name: $0, indent: indent + 2) }`
3. Change `sort(children)` calls to `children.sort()`

That should eliminate some extra copies/allocations from the `map`s.

rjchatfield:
That's a good idea. I just gave it a shot and sadly the test results failed. ☹️ Passing in the `name`  and the `indent` probably changes the sort order. Maybe that's a good breaking change to make?

Do you think this "good enough" non-breaking PR?

stephencelis:
Dang, yeah this is totally great as is. We can investigate further optimizations in the future!

rjchatfield:
For what it's worth, your idea is much quicker. 

And thanks for building a great library!

PR #199, opened 2019-03-25
Inline Snapshot Testing

rjchatfield:
This PR aims to add a new string literal-based API:

```swift
assertInlineSnapshot(of: user, as: .dump, toMatch: """
▿ User
  - bio: "Blobbed around the world."
  - id: 1
  - name: "Blobby"
""")
```

To record, set `record = true` or provide an empty string:

```swift
assertInlineSnapshot(of: user, as: .dump, toMatch: """
""")
// or
assertInlineSnapshot(of: user, as: .dump, toMatch: "")
```

Recording magic in action:

slide_for_effect

I've done a [talk](http://vimeo.com/326120139) and written a [blog post](https://medium.com/@rjchatfield/inline-snapshot-testing-e5e237fc47f3) on the idea. And I have plenty more to share. ;) 

Happy for feedback.


stephencelis:
This is awesome! The reason the build is failing is because we currently have to share logic with a `SnapshotTestCase` class on Linux, so you'll need to duplicate things there. Hopefully with Swift 5 we will be able to stop worrying about this, though. Assuming we drop Swift 4 compatibility today we can rebase this branch and hopefully everything just passes!

stephencelis:
Alright, 1.4.0 is out and I merged it into this branch! I think @mbrandonw and I may tweak the API and some of the messaging to make it consistent with other errors, but assuming it goes green I think we'll merge first!

stephencelis:
Just wanted to let ya know we plan on taking a look at this more in depth tomorrow and will merge if all looks good!

mbrandonw:
Hey @rjchatfield! Stephen and I just discussed this IRL and we really like what you've got! We made a few small changes to the API names here https://github.com/pointfreeco/swift-snapshot-testing/pull/199/commits/ffc714c6c089748abf19c9c00d3eb6499fb3e98c, and then we moved the inline assert helpers to their own file here https://github.com/pointfreeco/swift-snapshot-testing/pull/199/commits/8f62af662c987486c5a5d35ce3ccc015e0443b10.

We also decided to underscore the inline assert helpers so that we can use them a bit in our production code bases before we make it fully public. The helpers are marked as `public`, but this way since they are underscored we can iterate a bit on the API in case we find some rough edges.

We're going to merge now, and we'll let you know of any changes we decide to make in the future, and hopefully we'll get the non-underscored versions in our 1.5 release!

Thanks again for this, it's really awesome and we had never even considered this type of snapshot testing before!

rjchatfield:
Brilliant. Thanks guys. I got busier as the week went on, so I didn’t get a chance to come back to this. 

One of the big things I wanted to highlight was how `record` works. I was always performing the diff under the hood, and if you pass then I always bail out there. So even if you’re recording, I wanted to avoid touching the source files and showing errors in your test when nothing had changed. This was definitely important when we rerun all tests in our codebase to see if anything changed. 

But it seems you merged it in. So that’s great. I have some other ideas for improvements, but I can make separate PRs. 

Thanks again for being open to the idea. Your library has really helped us, so I hope this helps you.

PR #200, opened 2019-03-26
Swift 5

stephencelis:
Swift 5 snapshot testing!


Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -1,256 +0,0 @@
-import XCTest
```
stephencelis: 👋

stephencelis:
I think the macOS snaps are failing for dark/light/mojave/beta reasons.

mbrandonw:
> I think the macOS snaps are failing for dark/light/mojave/beta reasons.

In the circle logs it seems to be a crash:

> Restarting after unexpected exit or crash in SnapshotTestingTests/testMixedViews(); summary will include totals from previous launches.

Makefile
```diff
@@ -15,7 +15,7 @@ test-ios:
 	set -o pipefail && \
 	xcodebuild test \
 		-scheme SnapshotTesting_iOS \
-		-destination platform="iOS Simulator,name=iPhone XR,OS=12.1" \
+		-destination platform="iOS Simulator,name=iPhone Xʀ,OS=12.2" \
```
mbrandonw: smh

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -65,9 +64,9 @@ public func assertSnapshots<Value, Format>(
   line: UInt = #line
   ) {
 
-  strategies.forEach { name, strategy in
+  try? strategies.forEach { name, strategy in
```
mbrandonw: should we make `assertSnapshots` throw? on this cause any `value: () -> Value` throws to be swallowed?

Sources/SnapshotTesting/SnapshotTestCase.swift
```diff
@@ -1,256 +1,4 @@
 import XCTest
 
-#if !os(Linux)
 @available(swift, obsoleted: 5.0, renamed: "XCTestCase", message: "Please use XCTestCase instead")
 public typealias SnapshotTestCase = XCTestCase
-#else
-
-/// An XCTest subclass that provides snaphot testing helpers.
-open class SnapshotTestCase: XCTestCase {
```
mbrandonw: 👋

mbrandonw:
are you sure you want to record the precision test?

image

I think we want to test that they match even though they are slightly different.

Wonder if we should start checking if the snapshot "matches" before recording, but that would definitely increase record times by quite a bit :/

PR #203, opened 2019-03-27
Add cURL Representation Comparison Strategy

### LINKED ISSUE #88: Add a curl snapshot strategy for URLRequest

Sherlouk:
Closes #88 

Solution is based off of [Alamofire's solution](https://github.com/Alamofire/Alamofire/blob/master/Source/Request.swift#L624-L697).

I welcome feedback and any changes you would like to have happen 😄 

Documentation has been updated with usage and example

### DIFF
--- a/Documentation/Available-Snapshot-Strategies.md
+++ b/Documentation/Available-Snapshot-Strategies.md
@@ -39,8 +39,8 @@ If you'd like to submit your own custom strategy, see [Contributing](../CONTRIBU
       - [`.image`](#image-8)
       - [`.recursiveDescription`](#recursivedescription-3)
   - [`URLRequest`](#urlrequest)
+      - [`.curl`](#curl)
       - [`.raw`](#raw)
-      - [`.curlRepresentation`](#curlRepresentation)
 
 ## Any
 
@@ -700,6 +700,28 @@ Records:
 
 **Platforms:** All
 
+### `.curl`
+
+A snapshot strategy for comparing requests based on a cURL representation.
+
+**Format:** `String`
+
+#### Example:
+
+``` swift
+assertSnapshot(matching: request, as: .curl)
+```
+
+Records:
+
+```
+curl \
+	-X POST \
+	-H "Accept: text/html" \
+	-d 'pricing[billing]=monthly&pricing[lane]=individual' \
+	"https://www.pointfree.co/subscribe"
+```
+
 ### `.raw`
 
 A snapshot strategy for comparing requests based on raw equality.
@@ -720,26 +742,3 @@ Cookie: pf_session={"userId":"1"}
 
 email=blob%40pointfree.co&name=Blob
 ```
-
-### `.curlRepresentation`
-
-A snapshot strategy for comparing requests based on a cURL representation.
-
-**Format:** `String`
-
-#### Example:
-
-``` swift
-assertSnapshot(matching: request, as: .curlRepresentation)
-```
-
-Records:
-
-```
-curl \
-	-X POST \
-	-H "Accept: text/html" \
-	-d 'pricing[billing]=monthly&pricing[lane]=individual' \
-	"https://www.pointfree.co/subscribe"
-```
-
--- a/Documentation/Available-Snapshot-Strategies.md
+++ b/Documentation/Available-Snapshot-Strategies.md
@@ -40,6 +40,7 @@ If you'd like to submit your own custom strategy, see [Contributing](../CONTRIBU
       - [`.recursiveDescription`](#recursivedescription-3)
   - [`URLRequest`](#urlrequest)
       - [`.raw`](#raw)
+      - [`.curlRepresentation`](#curlRepresentation)
 
 ## Any
 
@@ -719,3 +720,26 @@ Cookie: pf_session={"userId":"1"}
 
 email=blob%40pointfree.co&name=Blob
 ```
+
+### `.curlRepresentation`
+
+A snapshot strategy for comparing requests based on a cURL representation.
+
+**Format:** `String`
+
+#### Example:
+
+``` swift
+assertSnapshot(matching: request, as: .curlRepresentation)
+```
+
+Records:
+
+```
+curl \
+	-X POST \
+	-H "Accept: text/html" \
+	-d 'pricing[billing]=monthly&pricing[lane]=individual' \
+	"https://www.pointfree.co/subscribe"
+```
+
--- a/Documentation/Available-Snapshot-Strategies.md
+++ b/Documentation/Available-Snapshot-Strategies.md
@@ -716,9 +716,9 @@ Records:
 
 ```
 curl \
-	-X POST \
-	-H "Accept: text/html" \
-	-d 'pricing[billing]=monthly&pricing[lane]=individual' \
+	--request POST \
+	--header "Accept: text/html" \
+	--data 'pricing[billing]=monthly&pricing[lane]=individual' \
 	"https://www.pointfree.co/subscribe"
 ```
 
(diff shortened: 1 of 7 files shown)


Documentation/Available-Snapshot-Strategies.md
```diff
@@ -40,6 +40,7 @@ If you'd like to submit your own custom strategy, see [Contributing](../CONTRIBU
       - [`.recursiveDescription`](#recursivedescription-3)
   - [`URLRequest`](#urlrequest)
       - [`.raw`](#raw)
+      - [`.curlRepresentation`](#curlRepresentation)
```
stephencelis: Wanna move this before `.raw`? We don’t document it but our one go-to for organization is alphabetizing.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,39 @@ extension Snapshotting where Value == URLRequest, Format == String {
 
     return ([method] + headers + body).joined(separator: "\n")
   }
+  
+  /// A snapshot strategy for comparing requests based on a cURL representation.
+  public static let curlRepresentation = SimplySnapshotting.lines.pullback { (request: URLRequest) in
```
stephencelis: Let’s call this `curl`. Nice and simple 😄

Documentation/Available-Snapshot-Strategies.md
Sherlouk: Not a problem, this is done 😄

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
Sherlouk: Sure thing! Done

Sherlouk:
ahh monkey I moved the example, but didn't update the usage! Let me fix that quickly...

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,39 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
+    case "GET": break
+    case "HEAD": components.append("--head")
+    default: components.append("-X \(httpMethod)")
+    }
+
+    // Headers
+    if let headers = request.allHTTPHeaderFields {
+      for (field, value) in headers where field != "Cookie" {
+        let escapedValue = value.replacingOccurrences(of: "\"", with: "\\\"")
+        components.append("-H \"\(field): \(escapedValue)\"")
```
mbrandonw: Whatdya think about expanding the short argument names for their longer versions?

```suggestion
        components.append("--header \"\(field): \(escapedValue)\"")
```

Since snapshots don't need to be particularly compact it might be nice to have to be intimately familiar with all of the shorthand curl arguments.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,39 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
+      }
+    }
+
+    // Body
+    if let httpBodyData = request.httpBody, let httpBody = String(data: httpBodyData, encoding: .utf8) {
+      var escapedBody = httpBody.replacingOccurrences(of: "\\\"", with: "\\\\\"")
+      escapedBody = escapedBody.replacingOccurrences(of: "\"", with: "\\\"")
+      
+      components.append("-d '\(escapedBody)'")
```
mbrandonw: same here:

```suggestion
      components.append("--data '\(escapedBody)'")
```

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,39 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
+  public static let curl = SimplySnapshotting.lines.pullback { (request: URLRequest) in
+
+    var components = ["curl"]
+
+    // HTTP Method
+    let httpMethod = request.httpMethod!
+    switch httpMethod {
+    case "GET": break
+    case "HEAD": components.append("--head")
+    default: components.append("-X \(httpMethod)")
```
mbrandonw: and here too:

```suggestion
    default: components.append("--request \(httpMethod)")
```

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,39 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
+    }
+
+    // Headers
+    if let headers = request.allHTTPHeaderFields {
+      for (field, value) in headers where field != "Cookie" {
```
mbrandonw: It seems that alamofire omits the `"Cookie"` header because it uses a `URLSessionConfiguration` under the hood. We may need to pluck out the cookie header and add it to the curl request via something like:

```swift
if let cookie = request.allHTTPHeaderFields["Cookie"] {
  components.append("-b \"\(cookie)\"")
}
```

I believe that is correct, but haven't thought too much about it myself.

Sherlouk: I had realised this but had also noticed [another script](https://gist.github.com/shaps80/ba6a1e2d477af0383e8f19b87f53661d) also skipping this (without taking the one from the Cookie field)

Wasn't sure whether there was something I didn't know about cURL/Cookies in headers so decided to skip over this behaviour.

Happy to add this, but will validate the functionality first

Sherlouk:
Fair review on the expanding of argument names, improved readability and understanding is never a bad thing. I'll make that change!

Sherlouk:
Apologies for the merge request history, had to do some weird branching as my machine is not currently compatible with Xcode 10.2 (Swift 5).

Have made the requested changes 😄 Thanks for the reviews!

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -16,4 +16,44 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
+    if let httpBodyData = request.httpBody, let httpBody = String(data: httpBodyData, encoding: .utf8) {
+      var escapedBody = httpBody.replacingOccurrences(of: "\\\"", with: "\\\\\"")
+      escapedBody = escapedBody.replacingOccurrences(of: "\"", with: "\\\"")
+      
+      components.append("--data '\(escapedBody)'")
+    }
+    
+    // Cookies
+    if let cookie = request.allHTTPHeaderFields?["Cookie"] {
+      components.append("--cookie \"\(cookie)\"")
```
stephencelis: Thanks for bringing this back! I think `cookie` should be escaped, too, though, just like the other headers above.

Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testURLRequest.post-curl.txt
```diff
@@ -0,0 +1,6 @@
+curl \
+	--request POST \
+	--header "Accept: text/html" \
+	--data 'pricing[billing]=monthly&pricing[lane]=individual' \
+	--cookie "pf_session={"user_id":"0"}" \
```
stephencelis: Ah yup, see this snapshot. I believe these quotes should be escaped!

Sherlouk: 😬 Oops! Good job someone has eagle eyes! Sorry about that

PR #212, opened 2019-04-01
Add "recording" to inline snapshotting

stephencelis:
Something I noticed missing while using inline snapshotting: a convenient way to re-record a single assertion.

### DIFF
--- a/Sources/SnapshotTesting/AssertInlineSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertInlineSnapshot.swift
@@ -13,6 +13,7 @@ import XCTest
 /// - Parameters:
 ///   - value: A value to compare against a reference.
 ///   - snapshotting: A strategy for serializing, deserializing, and comparing values.
+///   - recording: Whether or not to record a new reference.
 ///   - timeout: The amount of time a snapshot must be generated in.
 ///   - reference: The expected output of snapshotting.
 ///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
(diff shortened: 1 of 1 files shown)


Sources/SnapshotTesting/AssertInlineSnapshot.swift
```diff
@@ -90,14 +96,13 @@ public func _verifyInlineSnapshot<Value>(
 
       let trimmedReference = reference.trimmingCharacters(in: .whitespacesAndNewlines)
 
-      /// Always perform diff, and return early on success!
+      // Always perform diff, and return early on success!
```
stephencelis: A bit of cleanup/normalization along the way.

Sources/SnapshotTesting/AssertInlineSnapshot.swift
```diff
@@ -13,13 +13,15 @@ import XCTest
...
 ///   - snapshotting: A strategy for serializing, deserializing, and comparing values.
+///   - recording: Whether or not to record a new reference.
 ///   - timeout: The amount of time a snapshot must be generated in.
 ///   - reference: The expected output of snapshotting.
 ///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
 ///   - line: The line number on which failure occurred. Defaults to the line number on which this function was called.
 public func _assertInlineSnapshot<Value>(
   matching value: @autoclosure () throws -> Value,
   as snapshotting: Snapshotting<Value, String>,
+  record recording: Bool = false,
```
Sherlouk: Do we need to add `record: recording` to the `_verifyInlineSnapshot` call in this method?

stephencelis: Ha, good catch!

PR #213, opened 2019-04-02
Enable users to add userInfo to failure attachments

Sherlouk:
Hey 👋 

#### Justification:
I'm working on a separate tool for creating nicely formatted reports for our snapshot testing but we've hit a bit of a blocker when it came to handling the failure attachments which are available in the `TestSummaries.plist` file Xcode generates.

Essentially the file has an array of activities (depending on how many comparisons it does within one test), and inside each of them several attachments (depending on the diffing strategy) but there's no easy way to map an attachment back to the comparison it was running.

For instance we run our snapshot tests against 5 different devices. In this example, they all fail and we get 5 sets of 3 images. I would like the ability for my tool to then be able to show the 3 images under a header for which device they came from - this information is not currently available.

#### Solution:
With this merge request I've added an optional user info dictionary to the `verifySnapshot` method, this is then added to every failure attachment created for that run. This then let's our app store a small bit of information such as the device name alongside the attachment(s)

I've also expanded the macOS NSImage diffing to have titles for their attachments (much like iOS' UIImage) as well as ensuring that the attachments are kept even if the user does not fail their test (not mandatory with this library)

The test is designed to always fail (but the test will pass to prevent CI from locking up) - this enables us to validate the userInfo is stored within the log file. Unfortunately I couldn't find a way to validate the userInfo is actually stored automatically.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -149,6 +149,7 @@ public func assertSnapshots<Value, Format>(
 ///   - recording: Whether or not to record a new reference.
 ///   - snapshotDirectory: Optional directory to save snapshots. By default snapshots will be saved in a directory with the same name as the test file, and that directory will sit inside a directory `__Snapshots__` that sits next to your test file.
 ///   - timeout: The amount of time a snapshot must be generated in.
+///   - attachmentUserInfo: A dictionary of information which is added to failure attachments.
 ///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
 ///   - testName: The name of the test in which failure occurred. Defaults to the function name of the test case in which this function was called.
 ///   - line: The line number on which failure occurred. Defaults to the line number on which this function was called.
@@ -160,6 +161,7 @@ public func verifySnapshot<Value, Format>(
   record recording: Bool = false,
   snapshotDirectory: String? = nil,
   timeout: TimeInterval = 5,
+  attachmentUserInfo: [AnyHashable: Any]? = nil,
   file: StaticString = #file,
   testName: String = #function,
   line: UInt = #line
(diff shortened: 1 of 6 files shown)


Sources/SnapshotTesting/Snapshotting/NSImage.swift
```diff
@@ -20,9 +20,18 @@ extension Diffing where Value == NSImage {
       let message = new.size == old.size
         ? "Newly-taken snapshot does not match reference"
         : "Newly-taken snapshot@\(new.size) does not match match reference@\(old.size)"
+      let oldAttachment = XCTAttachment(image: old)
+      oldAttachment.name = "reference"
+      oldAttachment.lifetime = .keepAlways
```
stephencelis: What the reason for this change? It seems like we'd only want to keep attachments around for failing tests and clean up after them when the tests pass, no?

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -160,6 +161,7 @@ public func verifySnapshot<Value, Format>(
   record recording: Bool = false,
   snapshotDirectory: String? = nil,
   timeout: TimeInterval = 5,
+  attachmentUserInfo: [AnyHashable: Any]? = nil,
```
stephencelis: I wonder if we want to provide some kind of callback instead so that the user info can be derived from the attachment and/or value itself? Not sure attachments are inspectable enough to make this worth it, though.

Sources/SnapshotTesting/Snapshotting/NSImage.swift
Sherlouk: I'm not super precious about this so if we think it's best to not have it then that's cool too but my reasons were:

1. If we don't have this then it's not really possibly to have a unit test which demonstrates the usage.

2. If a snapshot has failed, I feel like we should still report the failure regardless and leave the reporting decisions up to the end client.

    • Perhaps one example would be if we run tests on 10 devices, and only one device fails we choose to have it show as passed but our custom reporting tool is able to still see the diffs and show that in a way which makes sense

Sources/SnapshotTesting/AssertSnapshot.swift
Sherlouk: I had considered some sort of mutating callback to modify the attachment but thought the solution was a bit more complicated for not a lot of real extra benefit.

There's not that many parts of the attachment we'd want to otherwise update other than the user info and perhaps (in contrast to the other comment) the lifetime of it's availability.

Sherlouk:
Thanks for taking a look @stephencelis, appreciate this is a bit of a different request than others but feel it moves in the direction of unlocking capabilities/increasing flexibility for different user implementations such as our own.

Having said that if you think this can be achieved in a better way then I'm all ears! 😄

PR #214, opened 2019-04-04
Sort headers in curl strategy.

crayment:
The curl strategy is currently useless for requests that have more than one header because they come out in a random order. This change sorts the headers and also introduces a second header (`Content-Type`) to one of the tests to help detect this going forward.

I'm open to any insights on making a consistently failing test. Adding more headers would increase the likeliness but not guarantee it.

Somewhat related, the curl strategy doesn't play nice with the new inline snapshotting feature due to the `\` characters not being escaped. I think some escaping should be done on the strings before inlining them into the file. For example try adding this test (it will fail with the auto-inlined string):

```swift
_assertInlineSnapshot(matching: get, as: .curl, with: """
curl \
  --header "Accept: text/html" \
  --header "Content-Type: application/json" \
  --cookie "pf_session={}" \
  "https://www.pointfree.co/"
""")
```

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
+++ b/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
@@ -52,8 +52,8 @@ extension Snapshotting where Value == URLRequest, Format == String {
 
     // Headers
     if let headers = request.allHTTPHeaderFields {
-      for (field, value) in headers where field != "Cookie" {
-        let escapedValue = value.replacingOccurrences(of: "\"", with: "\\\"")
+      for field in headers.keys.sorted() where field != "Cookie" {
+        let escapedValue = headers[field]!.replacingOccurrences(of: "\"", with: "\\\"")
         components.append("--header \"\(field): \(escapedValue)\"")
       }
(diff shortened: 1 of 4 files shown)


Sherlouk:
Hey, I can only apologise for the header ordering issue - great catch and thanks for working out a fix for that!

With regards to the inline snapshotting, unfortunately they were built in parallel so I wasn't really able to test them together. Though your solution there does also sound reasonable in escaping certain characters!

stephencelis:
@crayment Swift 5 raw strings negate the need to escape things. _E.g._ this works just fine:

``` swift
_assertInlineSnapshot(matching: get, as: .curl, with: #"""
curl \
  --header "Accept: text/html" \
  --header "Content-Type: application/json" \
  --cookie "pf_session={}" \
  "https://www.pointfree.co/"
"""#)
```

Ideally we detect the level of "raw"-ness, though, and escape accordingly.

crayment:
@stephencelis Thanks for following up! I love that solution. Didn't expect it to work with raw string literals already but it seems to be working great.

PR #217, opened 2019-04-12
Allow dynamic size of views based on trait collection content sizes

MarianaMeireles:
This fix enables to snapshot views objects whose size depends on the trait collection's content size. It also enables to snapshot views objects that don't have the size defined.
I encountered this when trying to test a label with different trait collections and no frame defined.

The following changes include tests and don't break the existing ones.

Now snapshots can be created and compared for views that the size depends on the trait collection's content:
```swift
let label = UILabel()
label.font = .preferredFont(forTextStyle: .title1)    
label.adjustsFontForContentSizeCategory = true
label.text = "What's the point?"

assertSnapshot(matching: label, as: .image(traits: .init(preferredContentSizeCategory: .small)))
```


Sources/SnapshotTesting/Common/View.swift
```diff
@@ -678,6 +675,17 @@ func prepareView(
     )
   }
   add(traits: traits, viewController: viewController, to: window)
+
+  if size.width == 0 || size.height == 0 {
+    // Try to call sizeToFit() if the view still has invalid size
```
stephencelis: Ah, very nice!

Issue #176, opened 2019-02-20
Output failing tests in 'ImageDiff' folder


hollanderbart:
Hi @stephencelis,

Thanks for creating this snapshot testing framework!
I was comparing this framework with the other snapshot framework (https://github.com/uber/ios-snapshot-test-case)(FBSnapshotTestCase) to know which features are supported.

What I like about this framework is the API, in which you can easily specify the device (iPhoneSE or iPad), while for with the FBSnapshotTestCase you have to run the test again on a different device (e.g. it will take the screen size from the simulator instead of via the test itself).

What I do like about the FBSnapshotTestCase framework is that it will output a folder called 'ImageDiff', which contains all the failing snapshots including the reference image and an image highlighting the difference. With this framework you don't do that but output a git diff.
I believe that having an output folder containing the failing test results in a better developer experience.

Is it possible to also create such a feature where you can have an 'ImageDiff' folder containing the failing tests?

Thanks ✌️

stephencelis:
Hi @hollanderbart! We output the failing snapshot to a temporary directory which can be reconfigured using an environment variable: `SNAPSHOT_ARTIFACTS`. I believe this is a partial solution to your problem: partial because we only output the new, different snapshot, we don't save the diff.

The "git diff" output is actually a configurable graphical diff tool output. You can even set `SnapshotTesting.diffTool = "ksdiff"` to get a copy-pastable output from a failing test.

And if you're using Xcode, we produce that image-based diff you like as an XCTAttachment. You can view these attachments from the "Report navigator".

All of this could use some better documentation, and we should definitely save the diff image to the temporary directory, too, for inspection, so I'll keep this issue open for now. We'll try to add this feature in the future, but if you'd like to contribute, let us know and we'll try to give you a push in the right direction!

SpacyRicochet:
Something that would be interesting is if, at the end, you output a single copy-paste concatenated ksdiff command that you can put into a shell and immediately open all failure differences that way.

Not sure how feasible that is without manual script additions though.

(Listening to @hollanderbart speaking about this right now 😋)

hollanderbart:
Agreed with @SpacyRicochet. In the ideal situation, I would like to see an 'ImageDiff' folder containing the 3 images (reference / test / diff) and a small script that would launch the `ksdiff` command.
I will investigate a bit more, thanks to the comment of @stephencelis.

hollanderbart:
[...]

Issue #193, opened 2019-03-18
File Save Permissions Error


Sherlouk:
Hi 👋 

First off just wanted to say thanks for the project, it's been a breeze to setup and super refreshing! -- Good work!

My question is currently whether or not this library is suppose to support physical devices or if the intended use is limited to Simulators?

When running on a physical device I get the below error when `fileManager.createDirectory` is called. The same is not true when ran from a simulator (which works as expected, without errors).

```
Error Domain=NSCocoaErrorDomain Code=513 "You don’t have permission to save the file “CaseNameTests” in the folder “__Snapshots__”." UserInfo={NSFilePath=/Users/name/path/to/repo/ModuleTests/__Snapshots__/CaseNameTests, NSUnderlyingError=0x1c28585d0 {Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted"}}
```

This appears to be the same issue referenced [here](https://github.com/pointfreeco/swift-snapshot-testing/issues/176#issuecomment-467328614)

Thanks in advance 😄

stephencelis:
Hi @Sherlouk! We recently added `snapshotDirectory` to the base `verifySnapshotHelper`: https://github.com/pointfreeco/swift-snapshot-testing/blob/5ae8db2a9cad57a698883fdad9e40b72654fd3dc/Sources/SnapshotTesting/AssertSnapshot.swift#L161

This means you can write your own `assertSnapshot` helper to change the directory in which the images are saved:

https://github.com/pointfreeco/swift-snapshot-testing/blob/5ae8db2a9cad57a698883fdad9e40b72654fd3dc/Sources/SnapshotTesting/AssertSnapshot.swift#L117-L143

We know this is a bit hidden and that the ergonomics aren't great, so we hope to improve the experience in the future.

Sherlouk:
Apologies in the delay getting back to you @stephencelis - I've updated my references to assert with my own implementation loosely following the example you showed above using a different directory in the top-level of my projects file structure but unfortunately am getting the same error (just paths updated to a different location).

Again, this is only happening when ran from a device - working fine from Simulator 😞Is there something different happening in file permissions when running from device?

Sherlouk:
I'll add that I've just done a further investigation with `uber/ios-snapshot-test-case` and they're getting the exact same issue where on device it doesn't save images, and on simulator it does.

[...]

PR #220, opened 2019-04-19
Write snapshot artifacts to disk

mac-gallagher:
Hi all! 👋

#### Justification
Right now, there is no way to write all of the `reference`, `failure`, and `difference` files found as `XCTAttachment`s  to disk. This makes it difficult to fix snapshot tests that fail on CI but pass locally.

The `SNAPSHOT_ARTIFACTS` environment variable gets us a third of the way there by writing the `failure` snapshot to disk, but not the other two. This PR allows for all three to be written to disk via the `SNAPSHOT_ARTIFACTS` variable. 

This enhancement was discussed in https://github.com/pointfreeco/swift-snapshot-testing/issues/176.

#### Proposed Solution
This PR introduces the `SnapshotArtifact` struct which is used to create the `XCTAttachment`s as well as optionally write the artifacts to disk. The enum `ArtifactType` identifies the artifact as one of the following types: `reference`, `failure`, and `difference`.

The proposed file structure looks something like this:

   ```
   - SNAPSHOT_ARTIFACTS
     - MyClassTests
       - testFunctionImageFormat.1
         - testFunctionImageFormat.1_difference.png
         - testFunctionImageFormat.1_failure.png
         - testFunctionImageFormat.1_reference.png
       - testFunctionStringFormat.1
         - testFunctionStringFormat.1_difference.txt
   ```

Open to any and all suggestions! Thanks!


stephencelis:
Hi @mac-gallagher! Thanks so much for the pull request! This looks great and definitely addresses a problem with how things are currently handled. @mbrandonw and I will take a closer look at this soon and deliver any feedback we may have!

mluisbrown:
@mac-gallagher just an FYI that you can use https://github.com/TitouanVanBelle/XCTestHTMLReport to get an HTML version of the Xcode test log, which will include any attachments. The created HTML report can then be saved in your CI artefacts, to allow browsing of snapshot failures, with diffs, for when they fail on CI.

It's a really nice workaround, which we have been using.

mac-gallagher:
> @mac-gallagher just an FYI that you can use https://github.com/TitouanVanBelle/XCTestHTMLReport to get an HTML version of the Xcode test log, which will include any attachments. The created HTML report can then be saved in your CI artefacts, to allow browsing of snapshot failures, with diffs, for when they fail on CI.
> 
> It's a really nice workaround, which we have been using.

@mluisbrown thanks for the suggestion! I'll definitely give it a try 👍

mac-gallagher:
Hey @stephencelis, any update on this PR?

Sources/SnapshotTesting/Diffing.swift
```diff
@@ -10,7 +9,7 @@ public struct Diffing<Value> {
   public var fromData: (Data) -> Value
 
   /// Compares two values. If the values do not match, returns a failure message and artifacts describing the failure.
-  public var diff: (Value, Value) -> (String, [XCTAttachment])?
+  public var diff: (Value, Value) -> (String, [SnapshotArtifact])?
```
stephencelis: This interface should have a backwards-compatible computed `var` (if possible).

Sources/SnapshotTesting/Diffing.swift
```diff
@@ -25,10 +24,27 @@ public struct Diffing<Value> {
   public init(
     toData: @escaping (_ value: Value) -> Data,
     fromData: @escaping (_ data: Data) -> Value,
-    diff: @escaping (_ lhs: Value, _ rhs: Value) -> (String, [XCTAttachment])?
+    diff: @escaping (_ lhs: Value, _ rhs: Value) -> (String, [SnapshotArtifact])?
```
stephencelis: Let's also add a backwards-compatible overload for this initializer.

mac-gallagher:
Hey @stephencelis! Thanks for the suggestion - I just pushed a commit adding backwards compatibility. A few things:
- I added a new computed var `artifactDiff` instead of replacing the existing `diff` variable. I also added an overloaded initializer.
- All of the internal snapshot strategies now use the `artifactDiff` variable, so it should be easy to soft deprecate the old method if/when the time comes!
- It was a little tricky to handle the backwards compatibility in `AssertSnapshot` and `AssertInlineSnapshot` in a _clean_ way. The respective verify functions will check for the `artifactDiff` variable first, and then the `diff` variable when the former is `nil`. The initializers are set up in a way where exactly one of them will be `nil` in the case of a failure.

Again, I'm open to any suggestions! Thanks!

mac-gallagher:
Gentle ping :) @mbrandonw @stephencelis

MaxDesiatov:
@mbrandonw @stephencelis this would be a super-useful feature to be merged to `master`, any updates on this?

megs-seek:
Agreed, would be great to get this merged in!

ixrevo:
Hi everyone, with Xcode 11 you can get all snapshot artifacts from CI pretty easy. Check out this issue for more info https://github.com/pointfreeco/swift-snapshot-testing/issues/291

justinguo:
This would definitely be a nice addition for the framework and cannot wait for using it!

stephencelis:
@mac-gallagher Thanks for making those changes. This fell off our radar but we're going to take a look at your changes tomorrow. Hope to get things merged quickly as well!

carsten-wenderdel:
Hi @stephencelis I would also like to see this merged. Using the whole `xcresult` file as artifact (proposed above in #291)  has some issues, for example the file size is bigger than what our CI system allows.

TomHumphrey150:
@stephencelis any updates on this? 🙏 

I'm working on something similar to write the snapshots to disk in the location that XCPretty uses for screenshots in its own HTML report. Would only be a minor tweak if this merged 😬

andrewmarmion:
@stephencelis Is there any update on this? It would be such a great addition.

teameh:
This would be a great addition! @stephencelis anything we can do to help get this merged?

Jake-Prickett:
Echoing many of the voices on here, this would be a great addition that we would really benefit from - anything I can do to help move this forward @stephencelis? 

I did come across this [comment in a separate issue](https://github.com/pointfreeco/swift-snapshot-testing/issues/291#issuecomment-574661371) which describes archiving the xcresult file on CI. We went down this route initially, but for large projects the xcresult file size gets pretty big (~19MB) which isn't ideal. 

This approach would minimize the amount of CI resources consumed and provide clear insights into failures that occur on CI. 

Are there concerns with the implementation or approach taken? Could this be something enabled/disabled via an environment variable?

mac-gallagher:
Sorry for all the activity above - I struggled to rebase my changes because I didn't realize the base branch changed from `master` to `main`! Anyways, resolved all the merge conflicts.

However, it doesn't look like that environment variable `SNAPSHOT_ARTIFACTS` is working right now... I'm getting the error `You can’t save the file ___ because the volume ___ is read only.` on the main branch right now. This will need to be fixed before these changes go in. It looks like there was some discussion about this in #193

nunogoncalves:
Hello! Any news on this one? Thank you very much for this work!! :)

nunogoncalves:
Hello! @stephencelis  Any news on this one? We would really like to have this available too. :)  
Thank you very much for this work and the lib!! :)

PR #224, opened 2019-05-02
XCUIElement (UI tests) strategy

vaderdan:
This pull adds strategy that compares snapshots of the XCUIElement 

These are all possible attributes accordingly Apple documentation 
```
public protocol XCUIElementAttributes {

    /*! The accessibility identifier. */
    var identifier: String { get }

    /*! The frame of the element in the screen coordinate space. */
    var frame: CGRect { get }

    /*! The raw value attribute of the element. Depending on the element, the actual type can vary. */
    var value: Any? { get }

    /*! The title attribute of the element. */
    var title: String { get }

    /*! The label attribute of the element. */
    var label: String { get }

    /*! The type of the element. /seealso XCUIElementType. */
    var elementType: XCUIElement.ElementType { get }

    /*! Whether or not the element is enabled for user interaction. */
    var isEnabled: Bool { get }

    /*! The horizontal size class of the element. */
    var horizontalSizeClass: XCUIElement.SizeClass { get }

    /*! The vertical size class of the element. */
    var verticalSizeClass: XCUIElement.SizeClass { get }

    /*! The value that is displayed when the element has no value. */
    var placeholderValue: String? { get }

    /*! Whether or not the element is selected. */
    var isSelected: Bool { get }
}
```

I did 2 new targets (Empty single page app and ui testing target), because I cannot mock `XCUIElement` from unit tests 

I tested this only on iOS

### DIFF
--- a/Documentation/Available-Snapshot-Strategies.md
+++ b/Documentation/Available-Snapshot-Strategies.md
@@ -41,6 +41,8 @@ If you'd like to submit your own custom strategy, see [Contributing](../CONTRIBU
   - [`URLRequest`](#urlrequest)
       - [`.curl`](#curl)
       - [`.raw`](#raw)
+  - [`XCUIElement`](#xcuielement)
+      - [`.recursivedescription`](#recursivedescription-4)
 
 ## Any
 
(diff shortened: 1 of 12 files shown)


stephencelis:
Hi @vaderdan! Thanks so much for the PR! @mbrandonw is traveling but when he gets back we will take a look and discuss.

stephencelis:
We finally had a chance to meet and discuss this today. 😄 We think it looks great and would like to help you get it merged! The one request we'd like to make is to ensure that the test app is incorporated into the Xcode project file using XcodeGen. Currently, we generate the project file using the following manifest:

https://github.com/pointfreeco/swift-snapshot-testing/blob/master/project.yml

This means that if we were to run `xcodegen` it would blow away all of your changes. **project.yml** should be the source of truth for the test project that's used to run UI tests. Even better, it'd be nice to ensure that these tests run in continuous integration if possible.

If you'd like to tackle either of these problems, let us know! If not, could you remove the test app and UI tests from this PR? Then we can try to get around to an XcodeGen-based solution in the future.

stephencelis:
@vaderdan Just wanted to know if you're still interested in getting this merged? We understand if you don't have the time and are happy to label this with "help wanted" instead!

Issue #230, opened 2019-05-17
Inline Snapshots: Escape Strings


stephencelis:
A regular multiline literal currently records the following:

``` swift
_assertInlineSnapshot(matching: value, as: .json, with: """
[
  "[\"key\": \"value\"]"
]
""")
```

This will fail on repeat test runs because `\"` is interpreted by Swift as `"`. We should instead escape `\`s encountered:

``` swift
_assertInlineSnapshot(matching: value, as: .json, with: """
[
  "[\\"key\\": \\"value\\"]"
]
""")
```

_Or_ promote the literal to a raw literal:

``` swift
_assertInlineSnapshot(matching: value, as: .json, with: #"""
[
  "[\"key\": \"value\"]"
]
"""#)
```

There's probably _some_ paring trickiness around escaping raw literals, though we can cross those bridges later.

Related: #228

mbrandonw:
👍 to the raw literal style, I think that solves it nicely!

PR #231, opened 2019-05-17
Escaped characters detection in literals

ferranpujolcamins:
Followup of https://github.com/pointfreeco/swift-snapshot-testing/pull/228, Fixes https://github.com/pointfreeco/swift-snapshot-testing/issues/230

This PR fixes inline snapshot tests containing special characters. AssertInlineSnapshot now checks whether the snapshot contains special characters or not. If it does, escapes the snapshot literal appropriately. 

Known issue: AssertInlineSnapshot still searches only for "" and multiline string literals (it does not search #""" """# for example)


SnapshotTesting.xcodeproj/project.pbxproj
```diff
@@ -647,7 +671,11 @@
 			buildSettings = {
 				BUNDLE_LOADER = "$(TEST_HOST)";
 				INFOPLIST_FILE = Tests/Info.plist;
-				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				LD_RUNPATH_SEARCH_PATHS = (
```
stephencelis: We auto-create the Xcode project file using XcodeGen. Can you run `make xcodegen` from the project directory and recommit that file?

Sources/SnapshotTesting/Common/String+SpecialCharacters.swift
```diff
@@ -0,0 +1,61 @@
+//
```
stephencelis: Cool to remove this header comment? We delete them in all of the other files.

Tests/SnapshotTestingTests/String+SpecialCharactersTests.swift
```diff
@@ -0,0 +1,133 @@
+//
```
stephencelis: Would you please remove this header comment, too?

Sources/SnapshotTesting/Common/String+SpecialCharacters.swift
ferranpujolcamins: Sure!

SnapshotTesting.xcodeproj/project.pbxproj
ferranpujolcamins: Done!

ferranpujolcamins:
> Do you think you'll have time to address this soon?

I'm giving it a try

ferranpujolcamins:
> One interesting side effect is that it'll automatically insert more #s than are needed when re-recording

I managed to make it work. Not without pain! Substrings are tricky. Mutating self in `AssertInlineSnapshot.replaceFirstOccurrence` produced crazy results, I don't exactly know why.

stephencelis:
Hey @ferranpujolcamins! Looks like we're close but not _quite_ there yet. If you run the following test block:

``` swift
  func testAny() {
    struct User { let id: Int, name: String, bio: String }
    let user = User(id: 1, name: ##"Bl#""#obby"##, bio: "Blobbed around the world.")
    assertSnapshot(matching: user, as: .dump)
    _assertInlineSnapshot(matching: user, as: .dump, with: #####"""
    """#####)
  }
```

The leading `#####"""` correctly becomes `##"""`, but the trailing `"""#####` ends up appending an extra coupla `#`s to become `"""#######`. I think the problem is the use of `multiLineLiteralEndIndex` for replacement, which only matches against `"""`.

The logic you're building on is definitely the most complicated part of the lib, sorry! Let us know if we can help in any way.

ferranpujolcamins:
I see! This won't be an easy change after all. Let's do this:
I'll open other PRs with small refactors that will make our live easier. My goal is to be able to unit test the inline snapshot feature, so I will separate code transformation from file operartions.

After that, we can keep working on this one.

ferranpujolcamins:
Would be cool to add swift-prelude as dependency?

ferranpujolcamins:
Mmmm I see it depends on this package so we can't use it here :(

ferranpujolcamins:
Ok, I've rebased onto master.
I've added some tests. Your counter example is now working properly.
I still need to add some more, but I'd appreciate early feedback.

The tests work as follows (it's a little bit confusing):
The tests in `InlineSnapshotTests` assert the correctness of the source file modifications that the inline snapshot test machinery does. To do so, we snapshot test the modified source.

The recorded snapshots, are in turn well formed tests that are added to the test target.
* Thus, we check the source code after an inline snapshot is recorded is valid. 
* They assert that the recorded snapshot is escaped properly.
Please see comment at the end of `InlineSnapshotTests.swift`.

Tests/SnapshotTestingTests/InlineSnapshotTests.swift
```diff
@@ -4,70 +4,411 @@ import XCTest
...
+    var recordings: Recordings = [:]
+    let newSource = writeInlineSnapshot(
+      &recordings,
+      Context(sourceCode: source, diffable: diffable, fileName: "filename", lineIndex: 1)
+      ).sourceCode
+
+    assertSnapshot(source: newSource, diffable: diffable)
+  }
+
+  // TODO: add more tests like this with different amount of lines
```
stephencelis: Just wanna flag this before merging.

Tests/SnapshotTestingTests/InlineSnapshotTests.swift
```diff
@@ -4,70 +4,411 @@ import XCTest
...
+
+extension Snapshotting where Value == String, Format == String {
+  public static var swift: Snapshotting {
+    var snapshotting = Snapshotting(pathExtension: "txt", diffing: .lines)
+    snapshotting.pathExtension = "swift"
+    return snapshotting
+  }
+}
+
+// Class that is extended with the generated code to check that it builds.
```
stephencelis: Cool stuff!

ferranpujolcamins:
Yeah I just added a couple of tests. Should be ready to merge. If we find some corner case later let's add a regression test!

Thanks guys!

ferranpujolcamins:
Gentle ping ;)

stephencelis:
Beautiful work! This'll eliminate a major gotcha for folks. Thanks again for contributing!

ferranpujolcamins:
Thank you guys for working on this useful project!

PR #232, opened 2019-05-21
Separate file IO from code transformation in InlineSnapshots code

ferranpujolcamins:
This PR separates the inline snapshot logic into IO and pure code transformation. The code transformation needs the `recordings` struct, so I made the new function `writeInlineSnapshot`monadic. This is useful to write the tests, since we can easily express what code transformations would happen on a real test:
```
let testExecution = pure
  >>> writeSnapshot1
  >>> writeSnapshot2
```

I wanted to use `swift-prelude`, but since it depends on `swift-snapshot-testing` I couldn't. So I brought some goodies from there and just copied the code here. I hope we can fix this in the future.


stephencelis:
Hey @ferranpujolcamins, this is looking really promising! @mbrandonw and I won't have a chance to sit down and really look over this for a bit, though, so sorry for the delay. After a quick skim the thing that sticks out most is bringing in some stuff from Prelude. I think we'll want to minimize that surface area if possible (for example, we defined a minimal `Async` type rather than bringing in `Parallel` with all its operators). Is it possible to trim the files you brought in down to bare essentials (that is, what you used in the PR)?

ferranpujolcamins:
Hi @stephencelis

Yes I’ll try to reduce the ported code up to a minimum.

By the way, are you ok using the state monad here? At the end this is not Haskell and a function with inout parameters would also work. If you think it’s best to keep it simple and ditch the state monad please let me know.

stephencelis:
Hi @ferranpujolcamins. Sorry for the delay! Let's definitely keep it simple and use `inout` instead. I think it'll make things easier for future contributors.

ferranpujolcamins:
Done! I used some operators. Now the tests are simple but in case we want to add more complex tests the operators will come handy.

stephencelis:
Looking good! Just one last thing and I think we can merge.

>  I used some operators. Now the tests are simple but in case we want to add more complex tests the operators will come handy.

@mbrandonw and I are definitely fans of operators, but since we're not really using them outside of that one test, maybe we avoid them for now. I'm thinking an inline closure won't be so bad? And should make the code more immediately readable for those unfamiliar with composition operators.

Tests/SnapshotTestingTests/InlineSnapshotTests.swift
```diff
@@ -0,0 +1,84 @@
+//
```
stephencelis: If you have time, let's also remove this header comment 🧹😄

ferranpujolcamins:
Fixed!

stephencelis:
Can you resolve the merge conflict? I'll merge as soon as the build is green 🎉

ferranpujolcamins:
Done!

PR #236, opened 2019-06-05
Using URL.init(fileURLWithPath:isDirectory:)

kirillyakimovich:
Using init(fileURLWithPath:isDirectory:), which allows you to explicitly
specify whether the returned NSURL object represents a file or directory to avoid file system examination.

https://developer.apple.com/documentation/foundation/nsurl/1410301-init
Quote from the documentation:
"If path does not end with a slash, init(fileURLWithPath:) examines the file system to determine if path is a file or a directory."


kirillyakimovich:
Thank you for open sourcing this amazing library!

PR #245, opened 2019-08-17
Publicize assertInlineSnapshot

stephencelis:
The one caveat we've found to this API is that it breaks undo. I think we consider the advantages far outweigh this negative, though, and hopefully Xcode will get better undo support for writes that happen outside of Xcode in the future!


stephencelis:
Let's revisit this in the future after SwiftSyntax.

PR #246, opened 2019-08-17
SwiftUI View Tests

natemann:
This adds UIImage tests for SwiftUI Views through UIHostingController.


Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
```diff
@@ -0,0 +1,47 @@
+//
```
stephencelis: Would you please remove the auto-generated header comments?

.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
stephencelis: Would you please add `.swiftpm` to the repo root directory's `.gitignore` and remove this file from the PR?

Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
```diff
@@ -0,0 +1,47 @@
...
+//  File.swift
+//  
+//
+//  Created by Nathan Mann on 8/17/19.
+//
+
+import Foundation
+import SwiftUI
+
+@available(iOS 13.0, *)
```
stephencelis: Let's scope these things to all SwiftUI platforms.

```suggestion
@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```

Might even need a `macCatalyst 13.0` in there but not sure if that works.

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -676,6 +677,24 @@ final class SnapshotTestingTests: XCTestCase {
     }
     #endif
   }
+    
+  @available(iOS 13.0.0, *)
```
stephencelis:
```suggestion
  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```

Might need to re-run tests on various platforms.

Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  File.swift
+//  
+//
+//  Created by Nathan Mann on 8/17/19.
+//
+
+import Foundation
+import SwiftUI
```
stephencelis: This file needs to be wrapped in a compiler directive to retain compatibility with Linux.

Hopefully this does the trick:

``` swift
#if canImport(SwiftUI)
…
#endif
```

Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
stephencelis: (I've omitted watchOS because it's my understanding that you can't XCTest it. If that's changed, though, would love to add support!)

natemann:
@stephencelis, thanks for the feedback. I made the necessary changes. A few things to note:

1. It looks like iOS 13 has some subtle font changes. This caused quite a few tests to fail. I updated the tests. I can take these updates out if you would like the PR to be smaller.
2. I am currently not running macOS 10.15, so I am not able to add extensions for SwiftUI.View and test them for NSImage, so this PR only covers iOS and tvOS.

VojtaStavik:
I wonder if there's a way to have the snapshot sized to the "intrinsic" size of the view. It's strange to have all snapshots sized to the size of the screen *or* to have to always provide an explicit size.

VojtaStavik:
This is what I came up with after some hacking, @natemann. Maybe it will be useful:
```
        .pullback { view in
          guard size == nil && config.size == nil else {
            return UIHostingController(rootView: view)
          }
          // If size is not set explicitly use the ideal size.
          let hostingController = UIHostingController(rootView: view.fixedSize())

          let idealSize = hostingController.sizeThatFits(in: .init(width: 0, height: 0))
          hostingController.view.frame = CGRect(origin: .zero, size: idealSize)

          return hostingController
        }
```

Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
```diff
@@ -0,0 +1,60 @@
...
+    size: CGSize? = nil,
+    traits: UITraitCollection = .init()
+    )
+    -> Snapshotting {
+      return Snapshotting<UIViewController, UIImage>
+        .image(on: config,
+               precision: precision,
+               size: size,
+               traits: traits)
+        .pullback(UIHostingController.init(rootView:))
```
mbrandonw: loving this point-free™️  transformation 😄

mbrandonw:
> This is what I came up with after some experiments, @natemann. Maybe it will be useful:

This is very interesting, thanks for doing some research to figure out how to make it nicer!

We think we'll leave this out of the base snapshot strategy for the time being. User of the library that want this functionality can do `assertSnapshot(matching: view.fixedSize(), as: ...)`. In the future we may be able to improve this more, but for now we want to try to keep the strategy as simple as possible as we try to figure out the best way forward.

Thanks again!

natemann:
@mbrandonw, no problem. I figured it would not be pulled in until Xcode was out of beta and iOS 13 was out.

concertman:
Hi all. With everything public now I was wondering what the status of this was and if I could help move it along in anyway. 

Thanks!

concertman:
@natemann @mbrandonw @stephencelis hey all. I was just wondering if there's anything I can do to help get this over the line. We've started heavily using SwiftUI in our project and want to continue using snapshot testing. Thank you!!

ecerney:
Any updates on getting this merged? Looks like it's been approved just has some conflicts now @natemann

justinguo:
Can anyone fix the conflicts for this PR? This would be great since we are using SwiftUI now for the UI project. Thanks.

stephencelis:
Sorry about the delay! We've been quite busy with other things. If anyone can take the time to fix the conflicts we'd love to get this merged.

stephencelis:
Thanks for rebasing, and sorry for the delay! This PR is a long time coming :smile:

natemann:
@stephencelis no problem. I realize you have been very busy with other projects.

regexident:
This PR as merged will always render at full device size, afaict, unless one provides a fixed desired size. Most of the time I would want to snapshot a view at its the ideal size though.

As such I ended up using a slight modification which takes `layout: PreviewLayout` instead of
`size: CGSize` (where the latter can be expressed as an instance of the former and thus is more expressive).

This also mirrors the API of `PreviewProvider` much nicer, I think.

```swift
extension Snapshotting where Value: View, Format == UIImage {
    public static func image(
        on config: ViewImageConfig,
        precision: Float = 1,
        layout: PreviewLayout = .sizeThatFits,
    ) -> Snapshotting {
        var config = config
        let size: CGSize?

        switch layout {
        case .device:
            size = nil
        case .sizeThatFits:
            config.size = nil
            size = nil
        case let .fixed(width: width, height: height):
            config.size = nil
            size = CGSize(width: width, height: height)
        @unknown default:
            fatalError("Unknown layout \(layout).")
        }

        return Snapshotting<UIViewController, UIImage>.image(
            on: config,
            precision: precision,
            size: size,
            traits: traits
        ).pullback { view in
            let controller = UIHostingController.init(
                rootView: view
            )
            if size == nil && config.size == nil {
                let maxSize = CGSize(width: 0.0, height: 0.0)
                let size = controller.sizeThatFits(in: maxSize)
                controller.view.frame = CGRect(origin: .zero, size: size)
            }
            return controller
        }
    }

    public static var image: Snapshotting {
        self.image(on: .init())
    }
}
```

cc @stephencelis

stephencelis:
@regexident I'm still wading through the backlog but did you open a PR for this? Definitely looks like a good improvement!

regexident:
Not yet, but gimme a sec … 😅

regexident:
Here you go, @stephencelis: https://github.com/pointfreeco/swift-snapshot-testing/pull/331

PR #248, opened 2019-08-22
Use the same bytesPerRow value in contexts to speed up image diffing

mr-v:
Hi!

At Babylon we have a lot of snapshot tests, so I looked into improving their performance. I managed to do so in three areas: image comparison, generating diff image (#249), and recording (#250).

This PR contains changes for image comparison. I'll open PRs for other changes separately as they are independent and I'm not sure whether you'd like to integrate all of them.

---

I profiled image comparison and the most expensive calls were drawing to context and converting to `pngData`. I noticed that [first `memcmp` check](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L69) usually fails for me. I looked at similar code in FB/Uber library and noticed that it's [using minimum value](https://github.com/uber/ios-snapshot-test-case/blame/b168a2f1e278b42422ed5ff5fb79b44b0f1f5c4c/FBSnapshotTestCase/Categories/UIImage+Compare.m#L54-L55) of `bytesPerRow` instead of maximum and later uses it to create contexts for both images. I dug a little more to make sense of it - it turns out that, as some kind of optimization, images are sometimes padded to multiples of 64 bytes per row. It looks like this was the cause of that first `memcmp` failure, e.g. fresh snapshot had 3008 `bytesPerRow` while the one stored in the PNG only 3000. Once the minimal value was taken into consideration and used in all contexts setup, the tests still passed and sped up. Thanks to this change the first `memcmp` check is often enough - [`new.pngData()`](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L71) is never reached and there's no draw to [`newerContext`](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L74). 

To get some real world data I did measurements on our UI framework snapshot tests. Results in seconds.

| Scenario | Run 1  | Run 2  | Run 3  | Run 4  | Run 5  | Run 6  |  Min  | Max  | Average  |
|---|---|---|---|---|---|---|---|---|---|
| CGContexts with distinct `bytesPerRow`  | 59,778  |  58,992 |  58,676 | 58,658  | 58,604  | 58,562  | 58,562  |  59,778 |  **58,878** | 
|CGContext with shared `bytesPerRow`  | 38,381  | 38,877  | 38,156  | 37,978  | 38,346  | 38,620  | 37,978  | 38,877  | **38,393**  |
<br/>

Now those tests take about **35% less time.**

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -61,8 +61,9 @@ private func compare(_ old: UIImage, _ new: UIImage, precision: Float) -> Bool {
   guard newCgImage.height != 0 else { return false }
   guard oldCgImage.height == newCgImage.height else { return false }
   // Values between images may differ due to padding to multiple of 64 bytes per row,
-  // e.g. freshly taken view snapshot may differ from one stored as PNG.
-  // At this point we're sure that size of both images is the same, so we can go with minimal `bytesPerRow` value.
+  // because of that a freshly taken view snapshot may differ from one stored as PNG.
+  // At this point we're sure that size of both images is the same, so we can go with minimal `bytesPerRow` value
+  // and use it to create contexts.
   let minBytesPerRow = min(oldCgImage.bytesPerRow, newCgImage.bytesPerRow)
   let byteCount = minBytesPerRow * oldCgImage.height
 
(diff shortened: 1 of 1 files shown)


mbrandonw:
Wow, thanks for these PR's! And double thanks for splitting them up into 3 logical pieces!

This first one is a no-brainer and a great find. Stephen and I want to discuss a bit more for the other two, but we'll get back to you soon!

mbrandonw:
Hey! So sorry for the delay in reviewing this, but @stephencelis and I finally got around to taking a look, and running our Point-Free test suite with this change. It looks good, so we're merging now.

Thanks again!

PR #250, opened 2019-08-22
Don't diff the snapshots as part of recording

mr-v:
At Babylon we noticed that recording image snapshots was taking a pretty long time. I looked into it and it turns out that [diffing is also performed while recording](https://github.com/pointfreeco/swift-snapshot-testing/blob/e35f2ee842eff87873a497dadcdc06f474e94a7d/Sources/SnapshotTesting/AssertSnapshot.swift#L222-L225). It's a pretty expensive call. As far as I understand the diff is used to create part of error message, i.e. snapshot strategy defines `diffMessage` if there's a difference between reference and newly recorded snapshot.

For `UIView` snapshots which are verified with `image` strategy the messages are:

- ["Newly-taken snapshot does not match reference."](https://github.com/pointfreeco/swift-snapshot-testing/blob/4a97f4bdb0507e7e8f91abdebe68c0704583fe01/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L20-L22) when images differ.
- The default message ("Recorded snapshot: …") when images are the same.

My thinking here is that **at point of recording I’m probably not interested whether the new image is different**. Most likely  I want to replace the reference snapshot with a new one. Because of this **I propose to skip the diff which causes performance issues** and always use the default `"Recorded snapshot: …"` message. I think the performance improvement justifies having a little less precise message. I've only used library with images, so I might be missing something though.

I profiled recording of a single test which verifies 6 snapshots. In most common scenario when we record after making some changes the difference is huge. I also verified scenario in which we may want to re-record everything in one go even when there are no differences and it's still considerably faster. Results in seconds.

| Scenario  |  Run 1 | Run 2  | Run 3  | Run 4  | Run 5  | Run 6  | Average  |
|---|---|---|---|---|---|---|---|
| master: record when different  | 5,502  |  5,519 | 5,431  | 5,406  | 5,482  | 5,474  | **5,469**  | 
| master: record when no difference  | 0,670  |  0,669 | 0,676  | 0,670  | -  |  - | **0,671**  |
| PR: record and skip diffing  | 0,484  |  0,475 | 0,481  |  0,469 | 0,472  |  - |   **0,476**  |

Improvments: 

- record when different - **11,5x faster**,
- record when no difference - **29% faster**.


mbrandonw:
We like this change for improving the performance of recording, but we are a little worried about losing the ability to show a nice diff for when recording textual snapshots (we use this to kinda do "snapshot-driven testing" in which we will repeatedly run our tests looking to see how the diff changes).

We want to explore an alternative to this in which the snapshot strategy gets to decide how it does this logic. We'll let you know how it goes!

felixscheinost:
+1

I'm (ab-)using `verifySnapshots` for waiting for async views. My tests are pretty slow so I tried commenting out diffing in `UIImage.swift` and also noticed a ~10x speedup (~200 ms compared to nearly 2 s) when there is a difference.

I'd like `verifySnapshots` to have an option to specify whether diffs should be generated and attached to Xcode because I'm e.g. calling it every 0.5 s for 5 s so besides being very slow there are lots of duplicate diffs attached in Xcode. I only want the last call to `verifySnapshots` to generate and attach diffs.

mr-v:
>  but we are a little worried about losing the ability to show a nice diff for when recording textual snapshots

That was my hunch - that there's some workflow that I might not be familiar with. 👍 

> We want to explore an alternative to this in which the snapshot strategy gets to decide how it does this logic. We'll let you know how it goes!

Did you have a chance to explore this?

stephencelis:
We haven't had time to yet, unfortunately. If you'd like to sketch out a potential solution we could start workshopping it together, though! It would be additional configuration logic to introduce to the `Snapshotting` (or `Diffing`) type.

stephencelis:
Actually, let's just merge this as is. Chatted with @mbrandonw and we think it's best to focus on performance and we can re-introduce ergonomics later on.

mr-v:
Awesome, thank you!

PR #252, opened 2019-09-25
Dark mode support

fespinoza:
I'm unsure why #251 was closed, but given the nature of this library, allowing to test dark mode
variants seems a good fit for this library.

In terms of tests, there is an addition to the `testTableViewController` to check for dark mode.

In order to have a reference image for the `testTableViewController`, the project needs to be executed in Xcode 11 I assume.

Then I would await for your review to see if there is anything to change/fix to add this feature to
the library.

**NOTE**: which device was used to record the present reference snapshots?
Answer: iPhone Xr (found on circle ci)


fespinoza:
The problem with the tests can be either related of the device the test were originally taken from or than there is some margin changes in iOS 13. I'm not sure about this

fespinoza:
One thing i'm debugging is:

Dark mode screenshots don't wont correctly in a project I'm running this code on: https://github.com/finn-no/FinniversKit/pull/662. As explained in the pull request the snapshot in dark mode (with the right traits) looks different than running the demo app in dark mode in the simulator.

In the case of the snapshots some colors react to the dark mode settings, some others don't, I'm not sure why but I think when applying dark mode there is an animation happening and the snapshot may be taken before the changes finish(?) at least as an initial theory.

When I create a new project to test this, the dark mode implementation support seems to work.
if someone wants to check that out can use this repo https://github.com/fespinoza/SampleSnapshotDarkMode

fespinoza:
@stephencelis any feedback regarding my PR? if you don't plan to support dark more or you don't agree with the approach you can just say so

Issue #255, opened 2019-10-04
Error when installing with SPM


maxxx777:
1. Create a new empty project (Xcode 11.1)
2. Add Package dependency through Xcode
3. Build the project

getting error:
```
ld: warning: Could not find or use auto-linked framework 'XCTest'
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_XCTAttachment", referenced from:
      objc-class-ref in SnapshotTesting.o
  "_OBJC_CLASS_$_XCTWaiter", referenced from:
      objc-class-ref in SnapshotTesting.o
  "_OBJC_CLASS_$_XCTContext", referenced from:
      objc-class-ref in SnapshotTesting.o
  "_OBJC_CLASS_$_XCTestExpectation", referenced from:
      objc-class-ref in SnapshotTesting.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Is it possible to install the library with SPM currently?

maxxx777:
my fault: added `SnapshotTesting` to main target.

stephencelis:
Yeah. We should add a caveat to installation instructions, though, since this seems to be the default behavior of Xcode's integration.

PR #256, opened 2019-10-05
Xcode 11 installation instructions

stephencelis:
Should hopefully prevent confusion like #255.


README.md
```diff
@@ -153,16 +172,6 @@ target 'MyAppTests' do
 end
 ```
 
-### Swift Package Manager
```
stephencelis: Also moved Carthage and CocoaPods below the fold to start with first-party tooling.

PR #254, opened 2019-09-29
Import FoundationNetwork if available

f-meloni:
👋,
On Swift 5.1 on Linux, `swift-snapshot-testing` is not compiling https://travis-ci.org/danger/swift/jobs/589176028#L538
This is related to https://forums.swift.org/t/foundationnetworking/24769.
Conditionally importing `FoundationNetworking` should fix that.

PR #259, opened 2019-10-10
Support for swift:5.1 on linux

finestructure:
I've run into issues building the package with swift:5.1, throwing the following in a few places:

```
/package/Sources/SnapshotTesting/Snapshotting/URLRequest.swift:3:39: error: use of undeclared type 'URLRequest'
extension Snapshotting where Value == URLRequest, Format == String {
                                      ^~~~~~~~~~
...
```

I've tested it with these image variants and they all pass:

```
FROM norionomura/swift:swift-5.0-branch
FROM swift:5.0
FROM swift:5.1
```

So it could go in with either - or maybe it's worth running the tests for both 5.0 and 5.1?

Hope that helps!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM norionomura/swift:swift-5.0-branch
+FROM swift:5.1
 
 WORKDIR /package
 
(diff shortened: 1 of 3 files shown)


stephencelis:
Thanks for taking the time to PR! There was another one in the stream that got submitted slightly earlier, so we merged it: #254 

Hopefully this unblocks you!

PR #261, opened 2019-10-18
Fix ordering of VC snapshot setup logic

dannyhertz:
According to Apple [docs](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html), you should call `addChild` *before* adding the child VC's view to the parent VC's view. Also, when looping over multiple strategies of varying sizes, there could be issues where the target view controllers view is not laid out again properly in time (we were only forcing layout on the container `rootViewController` vs the actual target `viewController`).

Reordering the methods and explicitly forcing layout on both the child and root vc fixed several issues we were having with nested collection views not laying out properly because the ordering of the tests (small to big vs big to small screen size).

Here is an example of what was happening *before* this fix. You can see that the iPhoneX snap still has the stale layout from the iphone8 snap:

failure_2_903492A2-E73E-47AE-8E75-6BD93078D58B_failure

And here is what it *should* look like now:

reference_1_903492A2-E73E-47AE-8E75-6BD93078D58B


SnapshotTesting.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name = "SnapshotTesting"
-  s.version = "1.6.0"
+  s.version = "1.6.1"
```
stephencelis:
```suggestion
  s.version = "1.6.0"
```

Let's keep the version the same for now.

Sources/SnapshotTesting/Common/View.swift
```diff
@@ -754,12 +757,16 @@ private func add(traits: UITraitCollection, viewController: UIViewController, to
...
+
   rootViewController.beginAppearanceTransition(true, animated: false)
   rootViewController.endAppearanceTransition()
-  window.rootViewController = rootViewController
+
+  viewController.view.setNeedsLayout()
+  viewController.view.layoutIfNeeded()
+
   rootViewController.view.setNeedsLayout()
   rootViewController.view.layoutIfNeeded()
```
stephencelis: Is the root VC layout step still needed?

dannyhertz:
Having a hard time coming up with a dummy example/test to reproduce what I'm seeing in my project. Will keep trying!

dannyhertz:
I take it back, was able to create a unit test that fails. The autoclosure for the matching param was throwing me off. Will update this PR!

Sources/SnapshotTesting/Common/View.swift
dannyhertz: Doesn't seem like it, but feels like a safe thing to keep in?

PR #268, opened 2019-11-02
Add a wait strategy extension.

mbrandonw:
Adds a strategy extension to transform any existing strategy into one that waits before executing. This is useful for waiting for UI animations to happen.

### DIFF
--- a/Sources/SnapshotTesting/Extensions/Wait.swift
+++ b/Sources/SnapshotTesting/Extensions/Wait.swift
@@ -2,6 +2,10 @@ import Foundation
 import XCTest
 
 extension Snapshotting {
+  /// Transforms an existing snapshot strategy into one that waits for some amount of time before taking the snapshot. This can be useful for waiting for animations to complete or for UIKit events to finish (i.e. waiting for a UINavigationController to push a child onto the stack).
+  /// - Parameters:
+  ///   - duration: The amount of time to wait before taking the snapshot.
+  ///   - strategy: The snapshot to invoke after the specified amount of time has passed.
   public static func wait(
     for duration: TimeInterval,
     on strategy: Snapshotting
--- a/Sources/SnapshotTesting/Extensions/Wait.swift
+++ b/Sources/SnapshotTesting/Extensions/Wait.swift
@@ -1,0 +1,23 @@
+import Foundation
+import XCTest
+
+extension Snapshotting {
+  public static func wait(
+    for duration: TimeInterval,
+    on strategy: Snapshotting
+  ) -> Snapshotting {
+    return Snapshotting(
+      pathExtension: strategy.pathExtension,
+      diffing: strategy.diffing,
+      asyncSnapshot: { value in
+        Async { callback in
+          let expectation = XCTestExpectation(description: "Wait")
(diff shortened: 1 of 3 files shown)


stephencelis:
Looks good!

niorko:
This is great extension, however I have an issue with UIViewControllers containing child viewControllers.

Those child VCs have their data loading methods bounded to lifecycle methods that are asynchronous (using promisekit).
Resulting image won't contain desired views at specific state with those childviewcontrollers.

My naive approach was to do this trial and error mumbo jumbo:
```swift
extension SnapShotTests {

    /// Call this function to trigger viewDidAppear / viewWillAppear
    func setNeedsDisplay(in viewController: UIViewController, inChildren: Bool = false) {
        viewController.view.setNeedsDisplay()
        viewController.loadViewIfNeeded()
        let _ = viewController.view // triggers lifecycle methods to call
        viewController.beginAppearanceTransition(true, animated: false)
        viewController.endAppearanceTransition()
        
        if inChildren {
            viewController.children.forEach { child in
                self.setNeedsDisplay(in: child, inChildren: inChildren)
            }
        }
    }
    
    func getSnapshotName(for viewController: UIViewController) -> String {
        viewController.getTopViewController().snapshotDescription
    }
    
    /// Asserts that a given value matches a reference on disk.
    ///
    /// - Parameters:
    ///   - value: A value to compare against a reference.
    ///   - snapshotting: A strategy for serializing, deserializing, and comparing values.
    ///   - name: An optional description of the snapshot.
    ///   - recording: Whether or not to record a new reference.
    ///   - timeout: The amount of time a snapshot must be generated in.
    ///   - file: The file in which failure occurred. Defaults to the file name of the test case in which this function was called.
    ///   - line: The line number on which failure occurred. Defaults to the line number on which this function was called.
    public func assertVCSnapshot(
        matching value: @autoclosure () -> UIViewController,
        as snapshotting: Snapshotting<UIViewController, UIImage> = .wait(for: .shortTimeout, on: .default),
        setNeedsDisplayDelay: Bool = false,
        named name: String? = nil,
        record recording: Bool = false,
        file: StaticString = #file,
        testName: String = #function,
        line: UInt = #line
    ) {
        let viewController = value()
        
        setNeedsDisplay(in: viewController, inChildren: true)
        if setNeedsDisplayDelay {
            let expectation = XCTestExpectation(description: "wait")
            DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
                self.setNeedsDisplay(in: viewController, inChildren: true)
                DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
                    expectation.fulfill()
                }
            }
            wait(for: [expectation], timeout: .infinity)
            setNeedsDisplay(in: viewController, inChildren: true)
        }
        
        assertSnapshot(
            matching: viewController,
            as: snapshotting,
            named: name ?? getSnapshotName(for: viewController),
            record: recording,
            file: file,
            testName: testName,
            line: line
        )
    }
}
```

... and I don't like it, this is not a proper solution. Sure, the best way would be to get rid of asynchronicity, but we can't do that right now.

Those lifecycle methods of child viewcontrollers are called after the timeout, in `prepareView` method - that is why I am doing this monkey dance before calling `assertSnapshot`.

Could you guys help me with this somehow? Thanks

Issue #269, opened 2019-11-04
Compilation fails on Linux CI + Swift 5.1.1


ctreffs:
Hi pointfreeco team,

thank's for the awesome library!

I am currently snapshot testing on both macOS and Linux.
Sadly on the Linux CI runner compilation of the library fails.

I am using the following version:
From `Package.resolved`:
```json
"package": "SnapshotTesting",
"repositoryURL": "https://github.com/pointfreeco/swift-snapshot-testing.git",
"state": {
  "branch": null,
  "revision": "35aac0fbc74cdb0848923d42ddbdd3400c938f72",
  "version": "1.6.0"
  }
```
on [Gitlab-CI](https://docs.gitlab.com/ee/ci/) using docker and the [official Swift 5.1.1](https://hub.docker.com/_/swift) image:
```sh
Running with gitlab-runner 12.3.0 (a8a019e0)
  on docker-auto-scale 0277ea0f
Using Docker executor with image swift:5.1.1 ...
Pulling docker image swift:5.1.1 ...
Using docker image sha256:f302143d0aa294a793d9ffc54a5ced12aee33ed44b529fd0bcd31635265a8db1 for swift:5.1.1 ...
```

The (first) specific error I get is:
```swift
.../swift-snapshot-testing/Sources/SnapshotTesting/Snapshotting/URLRequest.swift:3:39: 
error: use of undeclared type 'URLRequest'
extension Snapshotting where Value == URLRequest, Format == String {
...
```

Thanks for investigating.

stephencelis:
This was fixed, thanks!

PR #270, opened 2019-11-04
Fix compilation failure on Linux + Swift 5.1+ - fixes issue #269

ctreffs:
This PR fixes a compilation bug that I raised in https://github.com/pointfreeco/swift-snapshot-testing/issues/269.

On Linux `Foundation` does not contain `URLRequest` anymore, 
but was moved to `FoundationNetworking`.
See: https://stackoverflow.com/a/58606520/6043526 or https://forums.swift.org/t/pitch-move-urlsession-to-new-foundationnetworking-module/14002/25.

This PR fixes the compilation error that was caused by this:
```swift
Sources/SnapshotTesting/Snapshotting/URLRequest.swift:3:39: error: use of undeclared type 'URLRequest'
```

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
+++ b/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
@@ -1,4 +1,7 @@
 import Foundation
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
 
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks for submitting this fix! I'm closing it in favor of one submitted earlier, #254, but thanks for taking the time to submit.

PR #273, opened 2019-11-21
Wrap URL Request data with double quote instead of single quotes

ldstreet:
When snapshotting a post URLRequest that has a body which has been created by Codable, the cURL snapshot wraps the data in single quotes like this:

```
curl \
        --request POST \
        --header "Accept: application/json" \
        --header "Content-Type: application/json" \
        --data '{\"name\":\"tammy134235345235\", \"salary\":0, \"age\":\"tammy133\"}' \
        "http://dummy.restapiexample.com/api/v1/create"
```

Copying this cURL into terminal produces an invalid JSON error because the JSON is escaped incorrectly. Single quoted strings do not require escaping double quotes. Double quotes do. So then, I propose to make this change to double quote the body data similar to how the cookies section is quoted. This would transform the snapshot to this:

```
curl \
        --request POST \
        --header "Accept: application/json" \
        --header "Content-Type: application/json" \
        --data "{\"name\":\"tammy134235345235\", \"salary\":0, \"age\":\"tammy133\"}" \
        "http://dummy.restapiexample.com/api/v1/create"
```

Alternatively, we could choose to not escape the internal quotes, but that would make the code inconsistent with how cookies are represented.


mbrandonw:
Nice, this makes sense! Could you also update the [test](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Tests/SnapshotTestingTests/SnapshotTestingTests.swift#L619) for this strategy to get CI passing?

Thanks!

Sherlouk:
Seems like a sensible change! Is it worth also adding a new test which covers a JSON payload example?

ldstreet:
Got the URLRequest test passing, but many other tests seems to be failing locally. Seems like just small pixel differences for most. Do they need to be updated for iOS 13?

mbrandonw:
@ldstreet don't worry about those screenshot tests for now. we need to update them for ios 13 and Catalina. I'll take another look at this today and then I think we can merge!

ldstreet:
No problem. Thanks for the prompt response/approval! Any plans on when this will be able to make it into a release?

mbrandonw:
@ldstreet yes we'll try to get a release out soon! thanks for reminding us

PR #274, opened 2019-11-28
Allow reference images to be added to test report, even if passed

Sherlouk:
Hey 👋 

We automatically generate test reports using [XCTestHTMLReport](https://github.com/TitouanVanBelle/XCTestHTMLReport) and distribute this to members of our team (people who may not necessarily have quick access to our repository), all as part of our app release process.

We wanted to add the ability to *always* add reference images (and string diffs) to the test report, even if the test passes. This is so we can visually verify what "success" actually means.

This is not a breaking change, and reports will only increase in size if users add `attachReferenceImages: true` to their assertion calls.

I've also added the unique "name" (which has already been passed in previously) to the attachment activity. This is because we have multiple assertions in a single test which means it was difficult to distinguish between the different activities. -- This will be added automatically for any use who provides a name.

For those interested, this is how it looks using the report tool linked above 😄 

image


Sherlouk:
Noticing a failure on setting `XCTAttachment.name` - is this a linux-only issue that I just need to wrap?

Sources/SnapshotTesting/Common/AttachmentActivity.swift
```diff
@@ -0,0 +1,37 @@
...
+) {
+  #if !os(Linux)
+    guard ProcessInfo.processInfo.environment.keys.contains("__XCODE_BUILT_PRODUCTS_DIR_PATHS") else {
+      return
+    }
+    var message = "Attached " + (result.isSuccessful ? "Reference" : "Failure Diff")
+    if let name = name {
+      message += " - \(name)"
+    }
+    XCTContext.runActivity(named: message) { activity in
```
Sherlouk: We shouldn't run the activity if no attachment is going to be added - this causes a "Attached Reference" message to appear in the report but with no image.

Sherlouk:
Confirmed this is now working for us. Good from my side 👍

mbrandonw:
Thanks for the PR, this sounds interesting!

Just so I understand, when you use this do you really add `attachReferenceImages: true` to all of your `assertSnapshot` invocations? Or do you create your own assert helper that always passes `true` for the argument?

I'm mostly wondering if this is something you tend to turn on for all tests, or do you really need to be able to opt in and out of it depending on the test.

Sherlouk:
Our test suite has a small wrapper over this library to help with running each test in a large number of configurations - so yeah we just have one place that we update.

Happy to move to using environment variables if you'd rather that? (I personally find them a little less readable, especially as they're not documented anywhere currently!)

Sherlouk:
Rebased to get it up to date 👍 If you'd like it moving to ENV vars just give me a shout and I can make that change 😄

c0diq:
Would this be able to get merged soon?

Sherlouk:
I'd loved to have seen this merged, in fact we use it internally as our build since this is such an important feature to us!

If Brandon/Stephen confirm what changes they want I'm happy to make them to get it in 👍

PR #275, opened 2019-12-01
Update Package.swift, Circle, Travis for Swift 5.1

stephencelis:
Some extra work after #254 was merged.


stephencelis:
SceneKit and SpriteKit testing are unfortunately busted now due to unreliable rendering :/

I through too much time into it already, so opening an issue to track here: #276

Gonna merge, since this issue affects folks regardless of the changes made here.

PR #280, opened 2019-12-13
Prevent assertion of view with 0 width and or height

mackoj:
I don't think `swift-snapshot-testing` should assert when a view that it try to snapshot has 0 width, height or both. From my point of view it my be better to ignore it and create an artefact that represent it state.

A view can be in that state for a lot a justified reason(animation/state) and we might want to know that this view is actually with no width or height and it can change so being able to know that it is a 0 can actually be important.

Yes I could have used the `recursiveDescription` but it show way too many information so I made this PR.

This only support iOS I can do it for macOS too if needed.


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -240,6 +240,13 @@ public func verifySnapshot<Value, Format>(
       let data = try Data(contentsOf: snapshotFileUrl)
       let reference = snapshotting.diffing.fromData(data)
 
+      #if os(iOS)
```
stephencelis:
```suggestion
      #if os(iOS) || os(tvOS)
```

Sources/SnapshotTesting/Common/View.swift
```diff
@@ -628,6 +628,14 @@ extension View {
     #endif
     return nil
   }
+  #if os(iOS)
```
stephencelis:
```suggestion
  #if os(iOS) || os(tvOS)
```

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -704,6 +707,25 @@ final class SnapshotTestingTests: XCTestCase {
     }
     #endif
   }
+  
+  func testViewWithZeroHeightOrWidth() {
+    #if os(iOS)
```
stephencelis:
```suggestion
    #if os(iOS) || os(tvOS)
```

mackoj:
Thanks for the review @stephencelis I have done the modifications you suggested.

mackoj:
@stephencelis I know stewardship on open source project can be time consuming  be if you could find some time to merge this it would be great thanks. 🙏

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -32,6 +32,16 @@ extension Diffing where Value == UIImage {
       )
     }
   }
+  
+  
+  /// Used when the image size has no width or no height to generated the default empty image
+  private static func emptyImage() -> UIImage {
+    let label = UILabel(frame: CGRect(x: 0, y: 0, width: 200, height: 30))
+    label.backgroundColor = .red
+    label.text = "NO IMAGE GENERATED"
```
stephencelis, line 41: @mackoj What do you think of enhancing this error message with more information? Something like: 
```suggestion
    label.text = "Error: No image could be generated for this view as its size was zero. Please set an explicit size in the test."
```
If you don't have time to make the adjustments, that's fine! We can merge as-is and try to make the changes ourselves soon.

mackoj:
@stephencelis it's ready for re-review I have updated the text as you ask and regenerate fixture and update the label for the new text to fit in.

stephencelis:
Looks great! Thanks for taking the time 😄

PR #287, opened 2019-12-20
Fix Xcode project warning

maxx-cyber:


stephencelis:
:wave: Thanks for taking the time to submit a PR! We use [XcodeGen](https://www.github.com/yonaskolb/XcodeGen) to automatically generate our Xcode project file. Can you make the changes required there instead, and then run `swift run xcodegen` from the root of the repo to regenerate the project file? This may be a matter of upgrading the version of XcodeGen we use (defined in our SPM package description, here: https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Package.swift#L27-L33).

maxx-cyber:
Oh, sorry, I don't have expertise in XcodeGen. Could someone else do this?

stephencelis:
Just got around to doing this. Thanks for opening the issue 😄

PR #289, opened 2019-12-20
Don't overwrite test file on inline failure

### LINKED ISSUE #285: Inline Snapshots: Source Code is Deleted
In AssertInlineSnapshot.swift if we can't find a multiline literal then the entire Swift file is replaced with a string

> To use inline snapshots, please convert the "with" argument to a multi-line literal.

We should be a bit more careful with this... People (okay, me, in this example) may not have the latest code in source control when they run the test and end up just deleting all their changes. Xcode history doesn't help to undo!

Perhaps an XCTFail would work better here?

stephencelis:
This fixes #285.


stephencelis:
I have to step out but this was easy enough to do a quick fix. The longterm fix is to update inline snaps to use SwiftSyntax.

Sherlouk:
Thanks for looking into this! 🙌

PR #293, opened 2020-02-05
Swift 5.2 Fix

stephencelis:
Strangely the project builds just fine without these changes, but some Point-Free episode code samples that include SnapshotTesting as an SPM package won't compile in Xcode 11.4 with a "Circular reference" error. Changing to a computed var fixes, though!


stephencelis:
Any clue why Circle and Travis integrations are still active? Not seeing how to remove...

mbrandonw:
I can remove

dmcrodrigues:
Not related with SPM itself, we are using CocoaPods and I have the same problem. But is already fixed ❤️

Issue #298, opened 2020-02-11
Wait extension missing from .pbxproj and thus missing when using Carthage


gemigkaka:
The wait strategy extension in Sources/SnapshotTesting/Extensions/Wait.swift is not added to the XCode project (project.pbxproj), which makes it missing for Carthage users.

gemigkaka:
Solved by #300

PR #299, opened 2020-02-11
Added Wait and WaitTests to project + target for Carthage compability

gemigkaka:
Fix for #298.


mbrandonw:
Hey @gemigkaka, thanks for the PR!

We use XcodeGen for automatically creating the project file, and so typically we would ask for you to use that, but I went ahead and did that here #300 because we also wanted to update the version of XcodeGen we are using.

Once that merges you should be good to go!

gemigkaka:
Thanks a lot, then I'll close this PR.

PR #301, opened 2020-02-12
Add UITableViewCell with traitCollection snapshoting test

dstranz:
Hello,

I figure out that starting from Xcode 11.3* traitCollection is not properly propagated to UITableViewCells during snapshot making.

That's a test that visualize the problem and could prevent us from regression in the future.

✅ Xcode 11.1
❌ Xcode 11.3
❌ Xcode 11.3.1
❌ Xcode 11.4 beta


stephencelis:
These kinds of things can definitely be a bit of a pain 🤕 We'd love to take a PR if anyone has ideas on how to fix!

dstranz:
@stephencelis I found a workaround for this. Now it works fine in Xcode 11.3.1.

Could you take a look?

dstranz:
I recorded all snapshot again in latest stable Xcode, because frames changed a bit between versions. Do you know what’s wrong with tests? Looks like Travis is waiting for travis.yml 😕

dstranz:
@stephencelis I've submitted this issue via Apple Feedback - we will see if it's intended.

stephencelis:
@dstranz Thanks! Given that this is a bug on Apple's side do you think this PR should be closed for now? If it's possible to "hack" a fix into the snapshot strategy, that'd be nice, but it can come in a different PR.

dstranz:
Hello @stephencelis, 

I see that you fixed CI and all tests passed. Which Xcode is used there to run tests?

stephencelis:
@dstranz We don't constrain it right now: https://github.com/pointfreeco/swift-snapshot-testing/blob/master/.github/workflows/ci.yml#L14

So it looks like it's using 11.4.1. We set up the tests and CI back on Circle, where new Xcode versions were released very late. GitHub actions has made this situation much nicer, so we should update our CI and coverage to be nicer now.

That said, would you agree that this PR should be closed rather than merged? It's nice to know about this gotcha and report the bug to Apple, but I think we should ideally fix this behavior with a workaround that somehow applies the traits to the cell content rather than document the bad behavior in our test suite.

dstranz:
@stephencelis Yes, that's up to You. You can close it.

I'm only think why this test passed on CI 🤔

stephencelis:
@dstranz OH! I see now! Maybe it's been fixed??

Going to close this now, but if you find that it's still a problem and come up with a solution to it, we'd happily take a PR to fix! 😄

Issue #307, opened 2020-03-04
viewDidAppear did not called on iOS 13


sidepelican:
on iOS 13, `UIViewController.viewDidAppear` not called. This is because the viewController is not cleaned correctly.

mkko:
This issue with missing `viewDidAppear` breaks up some tests where view controllers are animated. Any plans getting this merged, @stephencelis, @mbrandonw, someone else?

stephencelis:
Just merged #308 in!

PR #308, opened 2020-03-04
add UIViewController lifecycle test and dispose used viewController

sidepelican:
Issue: viewDidAppear did not called on iOS 13 #307

When a view prepare subviews in its `UIViewController.viewDidAppear`, it can't take correct screenshot because it's not called on iOS13.
To fix this, the view needs to clean after that used.

in case of `testAssertMultipleSnapshot.2.png`, scroll bar was remaining because of previous snapshot. Maybe in common case, scroll bar was not want.


Sources/SnapshotTesting/Common/View.swift
```diff
@@ -653,7 +653,7 @@ func prepareView(
   traits: UITraitCollection,
   view: UIView,
   viewController: UIViewController
-  ) {
+  ) -> () -> () {
```
stephencelis, line 656:
```suggestion
  ) -> () -> Void {
```

Sources/SnapshotTesting/Common/View.swift
```diff
@@ -735,7 +738,7 @@ func renderer(bounds: CGRect, for traits: UITraitCollection) -> UIGraphicsImageR
   return renderer
 }
 
-private func add(traits: UITraitCollection, viewController: UIViewController, to window: UIWindow) {
+private func add(traits: UITraitCollection, viewController: UIViewController, to window: UIWindow) -> () -> () {
```
stephencelis, line 741:
```suggestion
private func add(traits: UITraitCollection, viewController: UIViewController, to window: UIWindow) -> () -> Void {
```

stephencelis:
Thanks for PRing and sorry for the delay!

Issue #310, opened 2020-03-10
Snapshotting view hierarchy throws error `UIViewControllerHierarchyInconsistency`


GHMarcus:
We get an error after the update from `1.7.0` -> `1.7.2`
You can reproduce the issue with the attached project.
[SnapshottingHierarchyError.zip](https://github.com/pointfreeco/swift-snapshot-testing/files/4311661/SnapshottingHierarchyError.zip)

**Test Setup:**
We wanted to snapshot the view hierarchy with an `UITabBarController` as the `rootViewController` from the `window`.

**The complete error:**
```caught "UIViewControllerHierarchyInconsistency", "adding a root view controller <SnapshottingHierarchyError.TabBarController: 0x7fd277837200> as a child of view controller:<UIViewController: 0x7fd276d26a40>"```

**The issue?:**
I searched for the reason and found it in the `View.swift` file in the line `746`:

`rootViewController.addChild(viewController)`

In our case the given `viewController` is an `rootViewController` so it can't bet add as a child to another. By deleting this line everything works fine. 
Does this work as intended, or is it in this case a bug?
I will try to fix this issue and submit a PR in the next days. Maybe you can give me a hint. 

And finally...
Thanks for your great work, really like to work with this framework.

stephencelis:
This should be fixed by #316. Please reopen if the issue persists, though!

adamcarter93:
I'm still seeing this issue myself. Is there any update on it?

My current workaround is to take my child VC and call `removeFromParent()` before asserting snapshots.

PR #316, opened 2020-03-25
check if VC is Root before adding to Window.rootVC

hassfers:
This should fix Issue: 
https://github.com/pointfreeco/swift-snapshot-testing/issues/310


Sources/SnapshotTesting/Common/View.swift
```diff
@@ -736,34 +736,35 @@ func renderer(bounds: CGRect, for traits: UITraitCollection) -> UIGraphicsImageR
...
+    } else {
+      NSLayoutConstraint.activate([
+        viewController.view.topAnchor.constraint(equalTo: rootViewController.view.topAnchor),
+        viewController.view.bottomAnchor.constraint(equalTo: rootViewController.view.bottomAnchor),
+        viewController.view.leadingAnchor.constraint(equalTo: rootViewController.view.leadingAnchor),
+        viewController.view.trailingAnchor.constraint(equalTo: rootViewController.view.trailingAnchor),
       ])
+    }
+    rootViewController.addChild(viewController)
+    rootViewController.setOverrideTraitCollection(traits, forChild: viewController)
```
stephencelis, line 760: Should traits still be added to either VC? It might be surprising to pass traits to the strategy and not have them reflected.

hassfers:
Hello, 
yes you are right. I moved it the the end so the traits are always added to rootVC

stephencelis:
Thanks! Will merge on green 😄

PR #325, opened 2020-05-06
Pass traits when snapshotting recursiveDescription in UIViewController

erikpoort:
Hi,

I found that the traits that you could set in recursiveDescription of UIViewController weren't passed. My goal was to test the layout of labels with a preferred size category, which I use to test the fix. See commits if anything is unclear!

Cheers


erikpoort:
@stephencelis You don't seem too active in reviewing PR's last few months, is it worth waiting for a merge or should I link to my own fork?

Issue #324, opened 2020-05-05
Diffing.image(precision:) should be scale aware


moritzsternemann:
While working with snapshot-tests on a large scale, we resorted to lowering the snapshot resolution in some cases to improve performance. This was achieved by reducing the `displayScale` of the trait collection passed to the `Snapshotting.image()` method.

While the results are as expected, when the snapshot-test fails, the attached diff is incorrect.

### Example
Notice the *plus icon* in the corner, which has the correct size while the *i icon* is scaled up.
![difference_3_56070A52-D4ED-4B9D-B02B-696523A763C6](https://user-images.githubusercontent.com/3034168/81066457-66b0c980-8edd-11ea-9dba-ed02832e0f09.png)

### Solution
I noticed that the image diffing strategy uses the `UIScreen.main.scale` constant to load the reference snapshot from disk ([Sources/SnapshotTesting/Snapshotting/UIImage.swift#L16](https://github.com/pointfreeco/swift-snapshot-testing/blob/master/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L16)).
This causes the reference image being a different size the new snapshot when creating the diff, which results in a pretty much useless XCTestAttachment.

I suggest to resolve this by passing the `displayScale` from the trait collection through to the image diffing strategy. I prepared the necessary changes and would like to open a PR as soon as possible if nobody finds anything wrong with my approach.

stephencelis:
This sounds good to us! Thanks for exploring the solution.

stephencelis:
@moritzsternemann This also appears to be a dupe of #243, right? I'm going to close this, but feel free to reopen if you think the issues are distinct.

moritzsternemann:
@stephencelis yes, I totally missed that!

Issue #243, opened 2019-07-28
Parameterizing `scale` of image strategies


nevillco:
I'm really enjoying adding Snapshot testing to a side project of mine. So far I just have a dozen or so, all of the form: `assertSnapshot(matching: vc, as: .image(on: .iPhoneX))`, so I was surprised when I found all of my tests failing because I had switched the target device that was running the tests. The source is:

```swift
public static func image(precision: Float) -> Diffing {
    return Diffing(
        toData: { $0.pngData()! },
        fromData: { UIImage(data: $0, scale: UIScreen.main.scale)! }
        ...
    )
}
```

I would suggest updating the function signature to `public static func image(precision: Float, scale: CGFloat = UIScreen.main.scale) -> Diffing`, and allow `ViewImageConfig`s such as `iPhoneX` to pass their native scale. I just wanted to open an issue first to see if this is a change you agree with, since it would in theory break consumers' tests that already have snapshots recorded at the wrong device scale.

proxi:
+1 I've run into the same unexpected issue with switching sims.

hollanderbart:
From the README:
```
⚠️ Warning: Snapshots must be compared using a simulator with the same OS, device gamut, and scale as the simulator that originally took the reference to avoid discrepancies between images.
```

proxi:
True that, I was aware of this, but if scale could be overriden, just like nevillco said, so that `on: device` would run in `device`'s native scale, overriding it would be a more sane choice. It seems like `(on: iPhoneX)` producing an image not matching iPhoneX scale is never a desired behaviour.

If it's not possible to override scale, perhaps trying to use device config of different scale than current device should fail?

hollanderbart:
Better yet, have an option to override scale or use simulator scale.
e.g. `assertSnapshot(...., overrideScale: true)`
With the list of devices, we could provide the scale variable

grycyk:
any updates here or a PR? Happy to help if needed, but I think this would be a very good addition.

kirillsh:
Seems like parametrizing scale wouldn't work either. 
I tried that and rendering of same fonts is still slightly different between 2x and 3x devices. 
Seems like it is because rasterizing of fonts is slightly different in two cases.
I guess it is controlled by the selected simulator's runtime, either for 2x or 3x simulator and I didn't find a way to parametrize that in `UIGraphicsImageRenderer` which is doing the actual rendering of a layer into UIImage.

[...]

PR #336, opened 2020-05-11
Make image diffing strategies custom scale aware

moritzsternemann:
This PR resolves the issue described in #324 and partially in #243.

When overriding the scale of image snapshots by setting the `displayScale` property of the passed trait collection, failure diffs were not rendered correctly because the screens scale was used to interpret the snapshot stored on disk.

To resolve this issue, I introduced a new scale parameter to the image diffing strategies which falls back to the display scale if it's not set or the default *undefined* value `0.0` of the `UITraitCollection`.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -4,16 +4,24 @@ import XCTest
 
 extension Diffing where Value == UIImage {
   /// A pixel-diffing strategy for UIImage's which requires a 100% match.
-  public static let image = Diffing.image(precision: 1)
+  public static let image = Diffing.image(precision: 1, scale: nil)
 
   /// A pixel-diffing strategy for UIImage that allows customizing how precise the matching must be.
   ///
   /// - Parameter precision: A value between 0 and 1, where 1 means the images must match 100% of their pixels.
+  /// - Parameter scale: Scale to use when loading the reference image from disk. If `nil` or the `UITraitCollection`s default value of `0.0`, the screens scale is used.
   /// - Returns: A new diffing strategy.
-  public static func image(precision: Float) -> Diffing {
+  public static func image(precision: Float, scale: CGFloat?) -> Diffing {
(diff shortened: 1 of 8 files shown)


ffittschen:
@mbrandonw @stephencelis do you have an estimate when you will have time to review this PR? Due to this scaling issue the diffs are unusable for us at the moment.

rcarver:
@ffittschen Thanks for this! I merged it into a fork and it worked perfectly.

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -4,16 +4,24 @@ import XCTest
...
+    if let scale = scale, scale != 0.0 {
+      imageScale = scale
+    } else {
+        imageScale = UIScreen.main.scale
+    }
+
     return Diffing(
       toData: { $0.pngData() ?? emptyImage().pngData()! },
-      fromData: { UIImage(data: $0, scale: UIScreen.main.scale)! }
+      fromData: { UIImage(data: $0, scale: imageScale)! }
```
stephencelis, line 24: Looks like the default undefined value is now `nil` instead of `0`, right? In that case could you remove the logic above and instead change this single line to:

```swift
      fromData: { UIImage(data: $0, scale: scale ?? UIScreen.main.scale)! }
```

moritzsternemann: I introduced this logic to handle the case of a `UITraitCollection` without a `displayScale` set being passed in. With such a trait collection, the default value of the `displayScale` is `0.0` ([see here](https://developer.apple.com/documentation/uikit/uitraitcollection/1623519-displayscale)).

Let me know if I didn't understand your concern right 🙂

stephencelis: Ah yup that makes sense! Thanks!

moritzsternemann:
Hey @stephencelis, I just noticed this was merged into master, not into main. Should I open a new PR against main or are you going to handle getting this into the right branch?

Issue #223, opened 2019-04-28
Save screenshot with device name


sger:
Hello,

First of all great library.I'm using the combos function from the kickstarter project to loop through 2 arrays the first one contains and array of languages and the second an array of devices from the ViewImageConfig class.

For example:

```
combos(Language.languages, SnapshotTestingDevices.landscape).forEach { language, device in
      let vc = ViewController()
      assertSnapshot(matching: vc, as: .image(on: device))
      assertSnapshot(matching: vc, as: .recursiveDescription(on: device))
   })
}
```

For the generated screenshot i want to use the device name.For example for the following device

```
ViewImageConfig.iPhoneX(.landscape)
```

output

```
testMyFunc_iphoneX_landscape.png
```

is this possible?

stephencelis:
Hi @sger! You can use the `name` parameter on `assertSnapshot` as documented here:

https://github.com/pointfreeco/swift-snapshot-testing/blob/75eaeb1f5019831102a0db7eb0a47e7cc4698ac2/Sources/SnapshotTesting/AssertSnapshot.swift#L25

sger:
Hi @stephencelis I know about this parameter the thing I want to achieve here is how to get a string representation for the struct:

```
ViewImageConfig(safeArea: __C.UIEdgeInsets(top: 0.0, left: 44.0, bottom: 24.0, right: 44.0), size: Optional((896.0, 414.0)), traits: <UITraitCollection: 0x6000014f2980; _UITraitNameUserInterfaceIdiom = Phone, _UITraitNameHorizontalSizeClass = Regular, _UITraitNameVerticalSizeClass = Compact, _UITraitNameUserInterfaceLayoutDirection = 0, _UITraitNameForceTouchCapability = 1, _UITraitNamePreferredContentSizeCategory = UICTContentSizeCategoryM>)
```

which is an iphoneXr for example if the device is:

```
ViewImageConfig.iPhoneXr(.landscape)
```

something like the following it will be very suitable

```ViewImageConfig.iPhoneXr(.landscape).name``` returns ```iPhoneXr_landscape```

stephencelis:
@sger We don't have that now but are definitely open to it! Would you be interested in submitting a pull request to make `ViewImageConfig` (and its nested types) conform to `CustomStringConvertible`?

sger:
@stephencelis I'm not sure about the implementation I've tried the following and it works
[...]

PR #340, opened 2020-05-20
Added `name` property to ViewImageConfig (#223)

philippec:
Once the device structs are created (e.g. `.iPhone8`), there is no good way to get a short description of what they are (`iPhone8_Portrait`) by using an extension since so much of what makes it unique is part of the initializer.

This `.name` property gives a descriptive short name which is useful when creating snapshot file names or log statements, for instance:

    [.iPhone8, .iPadMini].forEach { device in
        print("Device name: \(device.name)")
    }
    // "Device name: iPhone8_Portrait"
    // "Device name: iPadMini_Landscape"

Two notes:

1. I added the `name` parameter at the end of the list of `init` parameters because it read better when initializing with a name to have the traits and the name next to each other. I wasn't sure if you wanted the parameters to be in alphabetical order or if it was just happenstance. Compare:

`return .init(safeArea: safeArea, size: size, traits: .iPhoneSe(orientation), name: "iPhoneSe_\(orientation.name)")`

and

`return .init(name: "iPhoneSe_\(orientation.name)", safeArea: safeArea, size: size, traits: .iPhoneSe(orientation))`

2. I did not make this a `CustomDebugStringConvertible` as suggested in #223 because it would be a breaking change and I did not feel like it was right for me to submit a breaking change for a small feature. Other people might depend on the verbose output of `.debugDescription` for their tests.

If you want me to update my PR for any reason I'm obviously open to that. 

Thanks for making this available.


pblondin:
That would be nice to have this PR accepted. 👍

philippec:
I re-did this PR on `main` rather than `master`see #542

PR #349, opened 2020-06-12
PDF snapshotting

regexident:
This PR adds support for the following:

## `assertSnapshot` for `CALayer`s & `UIView`s using `.pdf`.

In addition to the existing …

```swift
assertSnapshot(matching: view, as: .image(precision: 0.95))
```

… we now also allow for …

```swift
assertSnapshot(matching: view, as: .pdf(precision: 0.95))
```

### Rendering

The snapshots are rendered using `UIGraphicsPDFRenderer` (instead of `UIGraphicsImageRenderer`) and saved as `.pdf` file.
This will **render the view's contents as vector** info wherever possible, **falling back to bitmaps** where not.

### Diffing

The **diffing** is performed by rasterizing the PDFs to an image and using the existing `Diffing<Image>`.
As such `precision: Float` is supported (and behaves) in the same way as known from `assertSnapshot(matching: …, as: .image)`.

### Scaling

The big benefit of persisting snapshots as `.pdf` rather than `.png` in the way that a vector PDF files' filesize is invariant to the dimensions of the image. This advantage becomes particularly useful for large fullscreen snapshots with large regions of uniform color, produced by a simple fill operation.

Snapshotting a `CAGradientLayer` like this …

```swift
let gradientLayer = CAGradientLayer()
gradientLayer.frame = CGRect(origin: .zero, size; size)
gradientLayer.colors = [UIColor.red.cgColor, UIColor.yellow.cgColor]
assertSnapshot(matching: gradientLayer, as: .image, named: "image")
assertSnapshot(matching: gradientLayer, as: .pdf, named: "pdf")
```

… results in the following file sizes:

| Dimensions  | PNG     | PDF |
|-------------|---------|-----|
| 10×10px     | 0.9 kb   | 3 kb |
| 100×100px   | 14 kb    | 4 kb |
| 1000×1000px | 1,785 kb | 4 kb |

### Vector vs. Bitmap

Many views (such as `UIButton`) render to bitmaps, when using `UIGraphicsPDFRenderer`, instead of native PDF vector paths. As such the cost-benefit of PDF highly depends on the nature of the view being tested. In general however (unless dealing with very small images) snapshotting as PDF should not have any noticeable overhead to pure images.

### Performance overhead

Due to the fact that …

```swift
assertSnapshot(matching: view, as: .pdf(precision: 0.95))
```

needs to perform the following chain of computations:

```
UIView -(UIGraphicsPDFRenderer)-> Data -(init)-> PDFDocument -(UIGraphicsImageRenderer)-> UIImage
```

… while …

```swift
assertSnapshot(matching: view, as: .image(precision: 0.95))
```

… needs just …

```
UIView  -(UIGraphicsImageRenderer)-> UIImage
```

… there is a certain performance overhead to be expected for the latter.

_You essentially end up trading space complexity for time complexity._

## `assertSnapshot` for `PDFKit`'s `PDFDocument`.

PDF documents can be diffed by comparing its rasterized pages based on pixel equality:

```swift
assertSnapshot(matching: pdf, as: .rasterized(precision: 0.95))
```

---

## Limitations

Since this PR uses `UIGraphicsPDFRenderer` (available in **iOS 10.0+**, **Mac Catalyst 13.0+** and **tvOS 10.0+**) and `PDFKit` (available in **iOS 11.0+** and **macOS 10.4+**) the API is limited to **iOS 11.0+**, only.

By making use of the underlying `CGPDFDocument` (of `CoreGraphics`) instead the number of supported platforms could be widened (adding `tvOS` and `Mac Catalyst`).

By using `func dataWithPDF(inside rect: NSRect) -> Data` ([docs)](https://developer.apple.com/documentation/appkit/nsview/1483797-datawithpdf#) on macOS instead of `UIGraphicsPDFRenderer` one could further add support for macOS.

However since I don't have any need for `tvOS` and `Mac Catalyst` myself right now and since such support could be added in a transparent and non-breaking way at a later point (and the equivalent `CoreGraphics` code being rather gnarly, requiring lots of additional boilerplate) I chose to make us of `PDFKit` instead, for now.


regexident:
Snapshotting via `.pdf` could possibly provide a solution for some of the issues discussed in https://github.com/pointfreeco/swift-snapshot-testing/pull/225.

stephencelis:
@regexident This looks very interesting. Thanks for the PR! We'll try to take a look at this soon.

regexident:
It might be a good idea to look into refactoring `func snapshotViewAsImage` & `func snapshotViewAsPDF` into a visitor-pattern to provide a convenient customization entry-point without needing to copy & paste several large functions.

I could think of several additional strategies that could benefit from it. Imagine for example a mode where you're generating a "wireframe" PDF of a view consisting of only solidly filled rectangles. This would provide a way to snapshot a view's layout at minimal storage use. Just one of many possibilities.

regexident:
Oops, didn't mean to close this PR.

PR #352, opened 2020-06-12
Fixed compilation when building with APPLICATION_EXTENSION_API_ONLY

### LINKED ISSUE #351: SnapshotTesting only builds from test targets with host application
As of version `1.7.2` and `1.8`, the library fails to build with in `View.swift`'s `prepareView`:
```
'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.
```

There's a `fatalError` with a runtime check, but that would never fail if it builds in the first place.

NachoSoto:
Fixes #351.

Unfortunately this code path isn't hit on tests so I couldn't verify if it works.
I considered other approaches, such as using `@available(iOSApplicationExtension, unavailable)`, but that would have required a major refactor because the `drawHierarchyInKeyWindow` comes from several layers up.

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -664,6 +664,18 @@ private final class NavigationDelegate: NSObject, WKNavigationDelegate {
 #endif
 
 #if os(iOS) || os(tvOS)
+extension UIApplication {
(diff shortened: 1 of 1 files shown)


NachoSoto:
Thank you!

PR #353, opened 2020-06-12
Changed project to APPLICATION_EXTENSION_API_ONLY = YES

NachoSoto:
The framework still builds and tests still pass, so I think this is better to ensure that the library continues to work with the more restrictive option.
This is a follow up to #352 and #351.

### DIFF
--- a/SnapshotTesting.xcodeproj/project.pbxproj
+++ b/SnapshotTesting.xcodeproj/project.pbxproj
@@ -950,6 +950,7 @@
 		0599123409D3A74965DD7DF5 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @NachoSoto! Thanks for another PR. We use [XcodeGen](https://www.github.com/yonaskolb/XcodeGen) to automatically generate our Xcode project file;

https://github.com/pointfreeco/swift-snapshot-testing/blob/master/project.yml#L14-L17

Can you make the changes there instead, and then run `swift run xcodegen` from the root of the repo to regenerate the project file? If not no big deal. I can make the change in a bit.

NachoSoto:
Ooooh of course, I'll do that :)

NachoSoto:
Done!

Issue #350, opened 2020-06-12
CocoaPods fails to validate SnapshotTesting 1.8.0+


stephencelis:
We just released 1.8.0, which includes support for SwiftUI. Unfortunately, we can't push to CocoaPods trunk due to validation errors:

```
 -> SnapshotTesting (1.8.0)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | xcodebuild:  note: Building targets in parallel
    - NOTE  | xcodebuild:  note: Planning build
    - NOTE  | xcodebuild:  note: Constructing build description
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift:19:45: error: no type named 'View' in module 'SwiftUI'
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SceneKit.swift:35:12: error: ambiguous use of 'image'
    - NOTE  | xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift:33:22: note: found this candidate
    - NOTE  | xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/UIView.swift:17:22: note: found this candidate
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SpriteKit.swift:35:12: error: ambiguous use of 'image'
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift:60:24: error: use of unresolved identifier 'UIHostingController'
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift:64:35: error: use of unresolved identifier 'UIHostingController'
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/UIView.swift:7:13: error: ambiguous use of 'image'
    - ERROR | [iOS] xcodebuild:  SnapshotTesting/Sources/SnapshotTesting/Snapshotting/UIViewController.swift:7:13: error: ambiguous use of 'image'
[...]

PR #355, opened 2020-06-13
Set minimum version to iOS 11.0 to fix issues with weak linking of SwiftUI

thedavidharris:
Seems to fix https://github.com/pointfreeco/swift-snapshot-testing/issues/350. Can verify by just doing an Xcode archive on this. Did not see any need to make changes to the tvOS or macOS targets, those seems to remain okay.


stephencelis:
@thedavidharris Unfortunately I'm still getting those same errors when I run `pod lib lint` on your branch ☹️

thedavidharris:
@stephencelis forgot to update the Podspec. Not on a machine I can validate it, but I think that would be why `pod lib lint` didn't pick it up.

thedavidharris:
Yeah, just checked, this does fix `pod lib lint` too. It's a weird issue, although given the low use of iOS 10 _and_ the fact that this is a test library, should be fine to bump it here.

PR #356, opened 2020-06-14
Add AccessibilitySnapshotPointfree Plugin

Sherlouk:
Created a hopefully helpful extension to AccessibilitySnapshot, so a quick README update to add it to the list!

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,8 @@ end
   - [swift-html](https://github.com/pointfreeco/swift-html) is a Swift DSL for type-safe, extensible, and transformable HTML documents and includes an `HtmlSnapshotTesting` module to snapshot test its HTML documents.
   
   - [GRDBSnapshotTesting](https://github.com/SebastianOsinski/GRDBSnapshotTesting) adds snapshot strategy for testing SQLite database migrations made with [GRDB](https://github.com/groue/GRDB.swift).
+  
+  - [AccessibilitySnapshotPointfree](https://github.com/Sherlouk/AccessibilitySnapshotPointfree) adds [AccessibilitySnapshot](https://github.com/cashapp/AccessibilitySnapshot) support for SnapshotTesting.
 
 Have you written your own SnapshotTesting plug-in? [Add it here](https://github.com/pointfreeco/swift-snapshot-testing/edit/master/README.md) and submit a pull request!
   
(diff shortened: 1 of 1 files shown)


stephencelis:
@Sherlouk Thanks for the PR! We saw Square's project and it's definitely very cool! We have a few thoughts and were wondering what you think.

- Under the hood, doesn't Square's project use (and depend on) iOSSnapshotTestCase? I wonder if there's a way to work with them to decouple this, since bringing your wrapper into a project would automatically bring in two different snapshot testing frameworks, which might be surprising 😅 If that's not possible, maybe we should add a caveat/warning to let folks know, both in your project's README and in our README where we would link to it?

- What do you think of naming the project "AccessibilitySnapshotTesting" instead of "AccessibilitySnapshotPointfree"?

Sherlouk:
Thanks for the response @stephencelis 

On the first point there - I've been working with the people over there to address this and we've actually worked together to break the iOSSnapshotTestCase dependency out into an optional subspec. My project only depends on their "Core" framework and therefore has no dependency on the Uber library.

On the second point, the naming was difficult - I had thought of "AccessibilitySnapshotTesting" but I didn't think it was immediately clear that it was relating to this tool and wasn't "different enough" to the CashApp library "AccessibilitySnapshot".

Sherlouk:
https://github.com/cashapp/AccessibilitySnapshot/issues/3 was the discussion, if you're interested.

stephencelis:
Oh very nice! Maybe on naming the pod we could use the category naming mentality of "AccessibilitySnapshot+SnapshotTesting"?

One other bummer is that AccessibilitySnapshot isn't SPM-compatible, do you can't provide an SPM module :( Maybe they'll address that in the future, but looks like they're still very Objective-C-heavy.

Sherlouk:
They're keen to adopt SPM support - one of their dependencies however does not which is the current blocker. Soon as they support it, I'll support it - though this is purely an optional add-on so I don't think that's a huge concern here!

I'll rename it to that, solid idea 👍

Sherlouk:
Just working on publishing the pod but should all be good 👍

Sherlouk:
Thanks for the recommendation there Stephen, think it's a much better name 👍 

and thanks for the merge, there's been a handful of people who have asked for this extension so that's props to good work over here!

mbrandonw:
This is really awesome @Sherlouk! Thanks for taking care of it and pushing for the necessary changes to `AccessibilitySnapshot` to make it a possibility!

PR #361, opened 2020-06-26
Fix Xcode 12 Name Clash (#360)

### LINKED ISSUE #360: Xcode 12 Beta name clash
In Xcode 12, there's a new method on `XCTestCase`: [`record(_ issue: XCTIssue)`][1].

Unfortunately, this clashes with SnaphotTesting's `record` boolean:

```swift
record = true
^^^^^^ Cannot assign to value: 'record' is a method
```

This can be worked around by prefixing the module name:

```swift
SnapshotTesting.record = true // Works fine
```

Still, I believe that it's unlikely that Apple will rename the new method until the release Xcode 12, so I suggest renaming the boolean `record`.

If you're interested, I'd love to submit a pull request.

[1]: https://developer.apple.com/documentation/xcode-release-notes/xcode-12-beta-release-notes

xavierLowmiller:
Fixes #360.

Replaces the `record` boolean with `isRecording` due to a name clash with Xcode 12's [record(_ issue: XCTIssue)][1] function.

This contains a proxy boolean for compatibility with existing codebases that is marked deprecated.
One could argue that this should only be deprecated in iOS 14, macOS 11, etc., but I think it should be deprecated right away so users can move to the new `isRecording` now and don't have breaking changes when migrating to Xcode 12.

[1]: https://developer.apple.com/documentation/xcode-release-notes/xcode-12-beta-release-notes

### DIFF
--- a/Sources/SnapshotTesting/AssertInlineSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertInlineSnapshot.swift
@@ -69,7 +69,7 @@ public func _verifyInlineSnapshot<Value>(
   )
   -> String? {
 
-    let recording = recording || record
+    let recording = recording || isRecording
 
     do {
       let tookSnapshot = XCTestExpectation(description: "Took snapshot")
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -8,6 +8,14 @@ public var diffTool: String? = nil
 /// Whether or not to record all new references.
 public var isRecording = false
 
+/// Whether or not to record all new references.
+/// Due to a name clash in Xcode 12, this has been renamed to `isRecording`.
+@available(*, deprecated, renamed: "isRecording")
+public var record: Bool {
+  get { isRecording }
+  set { isRecording = newValue }
+}
+
 /// Asserts that a given value matches a reference on disk.
 ///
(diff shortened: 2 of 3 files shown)


dstranz:
@xavierLowmiller  It's not better to deprecate global `record` property in favor of `SnapshotTesting.record`?

stephencelis:
@dstranz Unfortunately I don't think that's possible. Swift will always favor the `record` method defined on `XCTestCase` and there's no way to tell the compiler about the module variable.

dstranz:
> @dstranz Unfortunately I don't think that's possible. Swift will always favor the `record` method defined on `XCTestCase` and there's no way to tell the compiler about the module variable.

That's not such case?
https://github.com/pointfreeco/swift-snapshot-testing/pull/361/files#diff-d1c0c8ce08d2a1f90785bcfc178d2ebaR1070

stephencelis:
@xavierLowmiller If someone types `record = true` there is no way for us as a library to suggest a fix-it for `SnapshotTesting.record = true`, so it seems best to avoid a collision here.

xavierLowmiller:
Thanks for merging!

I just realized that I didn't update the Readme file, where it still says `record`.
Should I make another PR or can you just fix it on the main branch?

stephencelis:
@xavierLowmiller That'd be great, thanks!

PR #366, opened 2020-07-02
Fix crash when using `UIWindow` in `assertSnapshot`

zenangst:
`prepareView` in `View.swift` will now check if the current snapshot
target is a `UIWindow`. If that condition is `true`, then it will return
early and use the window as its dispose value. It will make the assumption
that the window has a `rootViewController` set and use that as it's `viewController`.

This avoids unbalanced calls to appearance transitions during test teardown
which will throw exceptions and crash the test.


zenangst:
@stephencelis mind taking a look a this? 😎

stephencelis:
Hey @zenangst! Thanks for taking the time to submit this and sorry if things are taking a bit longer to take a look at.

Given the early out does this mean that the traits and sizing will not be applied to the window?

zenangst:
Yeah traits and sizing should be left out. Maybe that is a false assumption on my part but if you give it the window then everything down the line should already be handled in your test. What do you think?

zenangst:
@stephencelis I pushed another commit to include the traits if the root view controller of the window could be resolved.
Is that an okay assignment seeing that `rootViewController` is optional in `UIWindow` and it is non-optional on the `add` method. Or should we fail early here and just throw an exception if the root view controller cannot be resolved?

Sources/SnapshotTesting/Common/View.swift
```diff
@@ -683,6 +683,21 @@ func prepareView(
...
+  ///
+  /// If we end up adding a window to another window, we can end up with unbalanced calls to
+  /// appearance transitions during test teardown.
+  ///
+  /// Example: `Unbalanced calls to begin/end appearance transitions for <UIViewController: *>`.
+  if let window = view as? UIWindow {
+    if let rootViewController = window.rootViewController {
+      return add(traits: traits, viewController: rootViewController, to: window)
+    } else {
+      return {}
```
zenangst, line 697: We could add an `assertFailure` here

zenangst:
ping @stephencelis 😎

zenangst:
Any headway here? Something that needs changing before it can be merged? 🌞

zenangst:
@stephencelis 

i3K1nMkuEtBeR

dmitrysimkin:
Please merge, I also encounter this issue

okankocyigit (approved):

okankocyigit:
@stephencelis will you merge this one, we are also having the same issue and it has been more then one year since @zenangst requested this PR?

That would be great if you merge, we don't want to continue with forked one, thanks :)

krzysztofpawski:
We have those changes on our fork for a while now and it works perfectly. Maybe this could be somehow merged? 🐱🙏
@stephencelis ? :)

zenangst:
giphy-47

zenangst:
waiting-waiting-patiently

PR #372, opened 2020-07-14
Fix sharedApplication selector in sharedIfAvailable (#365)

### LINKED ISSUE #365: Test crashes when 'drawHierarchyInKeyWindow' set to true
Snapshot test crashes with error `'drawHierarchyInKeyWindow' requires tests to be run in a host application` even though it is being run in a host application.

Test looks like this:
```swift
    func testViewController() throws {
        let vc = try XCTUnwrap(UIStoryboard(name: "Main", bundle: nil).instantiateInitialViewController())
        assertSnapshot(matching: vc, as: .wait(for: 0.5, on: .image(drawHierarchyInKeyWindow: true)))
    }
```

The problems seem to be in `UIApplication.sharedIfAvailable?` which returns `nil`.
Changing selector from `NSSelectorFromString("shared")` to `NSSelectorFromString("sharedApplication")` in `static var sharedIfAvailable: UIApplication? ` seems to fix the problem.

Is the something I'm missing in a test or a library setup (I've added it via SPM as said in a readme)?
Or could it possibly be a bug? I'll create a PR if that is a case.

mstultz:
Fixes #365 

The `sharedIfAvailable` property introduced in #352 should reference the Objective-C property name `sharedApplication`.

https://developer.apple.com/documentation/uikit/uiapplication/1622975-sharedapplication?language=objc

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -666,7 +666,7 @@ private final class NavigationDelegate: NSObject, WKNavigationDelegate {
 #if os(iOS) || os(tvOS)
 extension UIApplication {
     static var sharedIfAvailable: UIApplication? {
-      let sharedSelector = NSSelectorFromString("shared")
+      let sharedSelector = NSSelectorFromString("sharedApplication")
       guard UIApplication.responds(to: sharedSelector) else {
         return nil
       }


ldstreet:
Nice fix! Running into the same issue in my project. Any plans on a new release to add in this change?

stephencelis:
> Any plans on a new release to add in this change?

Just released 1.8.2 👍

PR #385, opened 2020-09-17
Introduce optional reference device checks

erikpoort:
Hi, I am part of a big team of many developers working a single app, where I introduced your library. Many times I will get one of these developers asking me why all the snapshots are failing, often the solution is to switch to the simulator which was used to create the reference images. They didn't know this was needed, and then they didn't know what simulator to use.

So this is my proposed solution. You can setup two vars, called referenceDevice and referenceMajorOSVersion. 
When these are set, for every snapshot (record on or off) the library will now check the simulator used, against these values.
If they do not match, the test will fail even before actually making the snapshot, explaining that you are using the wrong simulator.

I wanted to see what you think of it, before continuing to work on this!

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -8,6 +8,12 @@ public var diffTool: String? = nil
 /// Whether or not to record all new references.
 public var record = false
 
+/// Set this to the simulator used to create reference images for better failure messages
+public var referenceDevice: String?
+
+/// Set this to the OS version used to create reference images for better failure messages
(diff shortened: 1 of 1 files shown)


erikpoort:
@stephencelis @mbrandonw 
Hey guys, I was wondering what you though about this idea / approach / solution?

It's an optional and lightweight way to prevent developers from finding out about the simulator issue, before creating / committing / pushing the snapshots:
> ⚠️ Warning: Snapshots must be compared using a simulator with the same OS, device gamut, and scale as the simulator that originally took the reference to avoid discrepancies between images.

Happy to some more work based on this!

PR #387, opened 2020-09-18
Hash-based diffing of Hashable types

regexident:
While this implements support for hash-based diffing of `Hashable` types it does not resolve https://github.com/pointfreeco/swift-snapshot-testing/pull/225 quite yet (see discussion).

Even though hash-based view diffing is unlikely to produce satisfactory results (see discussion) being able to diff by hashes is useful enough on its own.


regexident:
Any chance to get this merged @stephencelis @mbrandonw?

mbrandonw:
Hi @regexident, thanks for the PRs! However, we think that instead of merging many snapshot strategies into the library itself that it may be more beneficial to set up a kind of community organization on GitHub where people can maintain their own strategies. We're in the beginning stages of setting this up (for both snapshot testing and TCA), but it will be a little longer.

For that reason we don't think we will merge this or #388 since they can both be supported without any internal changes to the library. We do have a [discussions](https://github.com/pointfreeco/swift-snapshot-testing/discussions) tab now open and so that could be a good place to discuss these kinds of experimental strategies with the users of the library.

regexident:
Hi @mbrandonw, no worries. To be honest transforming swift-snapshot-testing into a modular and extensible ecosystem within its own Github organization would be even better! I would be happy to join forces with you on this.

PR #389, opened 2020-09-20
Change the minimum ios deployment target to `10.0`

ibrahimkteish:


Sherlouk:
It might be worth checking out the PR it was bumped to iOS 11 as it was seemingly done for an important reason: https://github.com/pointfreeco/swift-snapshot-testing/pull/355

PR #391, opened 2020-09-25
Made `.data` snapshotting strategy `public`

regexident:
For snapshot testing binary artifacts being able to test against the bare data is preferable over having to snapshot it as `Encodable`, which then might even end up getting encoded as strings (for `.json` at least), making it impossible to inspect with a hex editor.


stephencelis (approved):
Ah yes, this should definitely be publicly available! Thanks!

PR #396, opened 2020-10-13
Make image diffing strategies custom scale aware

moritzsternemann:
*As the original PR with this code (#336) was opened before the migration of the default branch name from 'master' to 'main', it got merged into the old 'master' branch. I'm opening it again with exactly the same changes, just with the correct 'main' branch as the target.*

This PR resolves the issue described in #324 and partially in #243.

When overriding the scale of image snapshots by setting the `displayScale` property of the passed trait collection, failure diffs were not rendered correctly because the screens scale was used to interpret the snapshot stored on disk.

To resolve this issue, I introduced a new scale parameter to the image diffing strategies which falls back to the display scale if it's not set or the default *undefined* value `0.0` of the `UITraitCollection`.


moritzsternemann:
@stephencelis @mbrandonw It would be great if we could get this merged. It was approved and merged into the wrong base branch before: #336

pokryfka:
I think this is a very needed change.

However even though it helps to keep resolutions the renders are different when using different simulators.

attached examples of a snapshot of a view using iPhone 11 and iPhone 12 simulators.

gobetti:
@stephencelis you've merged #336 , so it should be fine to merge #396 right?

@pokryfka I understand this might not help everyone, but it does help my scenario and more people may have something similar. I have a few tests where I'm comparing images that my app generates (not app screenshots), and in unit tests I'm forcing those images to be generated at a 1x scale so different devices output the very same image. In the latest release of swift-snapshot-testing my tests already succeed using this strategy, the problem is when I introduce a change that fails the test: the `difference` image ends up being a mix of different scale images. This PR solves that.

dirtyhabits97:
Any updates on this?

As other people have pointed out #336 was merged to the wrong branch.

omarzl:
Hello @stephencelis!
Could you give us a hand with this pull request? We have an issue that can be solved with these changes 🙂

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -4,16 +4,24 @@ import XCTest
...
   /// - Parameter precision: A value between 0 and 1, where 1 means the images must match 100% of their pixels.
+  /// - Parameter scale: Scale to use when loading the reference image from disk. If `nil` or the `UITraitCollection`s default value of `0.0`, the screens scale is used.
   /// - Returns: A new diffing strategy.
-  public static func image(precision: Float) -> Diffing {
+  public static func image(precision: Float, scale: CGFloat?) -> Diffing {
+    let imageScale: CGFloat
+    if let scale = scale, scale != 0.0 {
+      imageScale = scale
+    } else {
+        imageScale = UIScreen.main.scale
```
stephencelis, line 19: Just making a quick indent change before merging. We haven't automated formatting in this repo yet 😅
```suggestion
      imageScale = UIScreen.main.scale
```

stephencelis (approved):
Sorry for letting this go by the wayside!

PR #397, opened 2020-10-15
when snapshotting a view controller, if its the root view controller, don't touch its window

littlebobert:


littlebobert:
Ultimately, using a per-pixel tolerance was too slow for us. Closing for now.

Issue #400, opened 2020-11-02
Snapshotting for XCUIElement


mackoj:
Hi,

It could be great to be able to do snapshotting of XCUIElement, have a better integration with XCUITesting in general.

This will simplify the use of Snapshot Testing with [XCUITests](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/testing_with_xcode/chapters/09-ui_testing.html).

```swift
    let app = XCUIApplication()
    app.launch()
    assertSnapshot(matching: app, as: .image)
```

SeanMcTex:
Interesting! I've always felt like one of the nice advantages of the library is that you don't have to drag all of the UI testing baggage with you when doing snapshot tests. What could you do with this that you can't do currently? (Not asking adversarially -- just trying to understand the benefits.)

mackoj:
Hi @SeanMcTex,

Thanks for the question.

Let say you already have UITests the way they work can't really guarantee that you don't break you UI.

With this you can add a new layer of validation to your already existing tests easily.

Snapshots have a lot of advantages that still applies to Xcode UITests.

We have non uitests with snapshots and we would like to used them in the uitests too.

fabiochoupina:
Hey @mackoj 

Not sure if it helps, but I've been able to import the SnapshotTesting framework and use the existing UI Tests structure to add a step to take a screenshot.

Here it is:
```swift
// Screenshot as UIImage
let screenshot = app.windows.firstMatch.screenshot().image
// Check against baseline
assertSnapshot(matching: screenshot, as: .image)
```

The issue is that the time on the phone is always different (making the screenshots always different), so to solve this, you could add a test pre-action script on Xcode with the following command:
```bash
xcrun simctl --set testing status_bar booted override --time "9:41"
```

mackoj:
@fabiochoupina I did a PR with a fix ;) https://github.com/pointfreeco/swift-snapshot-testing/pull/401

mbrandonw:
Hi @dr-star, see [this](https://github.com/pointfreeco/swift-snapshot-testing/issues/400#issuecomment-765060051) reply on how you can accomplish this.

And since this isn't really an issue with the library, and instead more of a feature request (which is better suited for [discussions](https://github.com/pointfreeco/swift-snapshot-testing/discussions)), I am going to close this.

PR #401, opened 2020-11-02
Add support for snapshotting XCUIElement

mackoj:
This add support for  XCUIElement - fix for #400


mackoj:
CI is fixed here #402

PR #402, opened 2020-11-02
Make CI work with Swift 5.3

mackoj:
CI seems to not be working anymore this is an attempt to fix it.

Pubgrub resolver is now enabled by default https://github.com/apple/swift-package-manager/pull/2421.


stephencelis:
Oh ha, thanks @mackoj! We need to go through some of these soon, sorry about the upstream fix!

mackoj:
@stephencelis no worries

Issue #333, opened 2020-05-07
There is no iPad 9.7" size


zlib:
SnapshotTesting.ViewImageConfig doesn't have such size.

stephencelis:
Hi @zlib! Mentioned in #332 but we'd love a contribution if you have time to add it here: https://github.com/pointfreeco/swift-snapshot-testing/blob/dcaeca0b80d2efdd881c67cdd62f7478fc5bec9f/Sources/SnapshotTesting/Common/View.swift#L15

stephencelis:
Fixed in #405.

PR #405, opened 2020-11-05
Add support for iPad 9_7 and iPad 10_2

skols85:
Adds support for 9'7 and 10'2 inch size iPads.
Fix for #333


stephencelis (approved):
Thanks so much for contributing these, and sorry for the delay! PR looks great.

PR #435, opened 2021-02-23
Speed up NSImage comparison

finestructure:
I've noticed that NSImage diffing is quite slow when using precision < 1 and large images. I guess that's mostly unavoidable but here's a little tweak that made it more than twice at fast in our case:

```
❯ swift compare.swift
false
compare1, precision 1    : 0.20623600482940674
true
compare1, precision 0.999: 5.742478966712952
false
compare2, precision 1    : 0.08092296123504639
true
compare2, precision 0.999: 2.4135000705718994
```

I'm attaching the test script and the two files I used below.

https://gist.github.com/finestructure/84b0b58b567e857b4f38c0e07882eecf

new
orig

(BTW: the tiny three pixel diff we're seeing in those images are between an M1 and an Intel Mac with otherwise identical OS and Xcode versions 🤪)

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/NSImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/NSImage.swift
@@ -77,12 +77,17 @@ private func compare(_ old: NSImage, _ new: NSImage, precision: Float) -> Bool {
   let newRep = NSBitmapImageRep(cgImage: newerCgImage)
   var differentPixelCount = 0
   let pixelCount = oldRep.pixelsWide * oldRep.pixelsHigh
-  let threshold = 1 - precision
-  for x in 0..<oldRep.pixelsWide {
-    for y in 0..<oldRep.pixelsHigh {
-      if oldRep.colorAt(x: x, y: y) != newRep.colorAt(x: x, y: y) { differentPixelCount += 1 }
-      if Float(differentPixelCount) / Float(pixelCount) > threshold { return false}
+  let threshold = (1 - precision) * Float(pixelCount)
+  let p1: UnsafeMutablePointer<UInt8> = oldRep.bitmapData!
+  let p2: UnsafeMutablePointer<UInt8> = newRep.bitmapData!
+  for offset in 0 ..< pixelCount {
+    if p1[offset] != p2[offset]
+        || p1[offset + 1] != p2[offset + 1]
+        || p1[offset + 2] != p2[offset + 2]
+        || p1[offset + 3] != p2[offset + 3] {
+        differentPixelCount += 1
     }
+    if Float(differentPixelCount) > threshold { return false }
   }
   return true
 }


finestructure:
PS: three tests are failing for me locally even without this change and I'm pretty sure it's also due to me running on an M1.

I see CI failed on macOS but can't see the details unfortunately.

stephencelis (approved):
This is great, thanks @finestructure! Don't worry about CI, which I think is pretty out-of-date, snapshot-wise. I'll merge and clean things up on the main branch.

Sources/SnapshotTesting/Snapshotting/NSImage.swift
```diff
@@ -77,12 +77,17 @@ private func compare(_ old: NSImage, _ new: NSImage, precision: Float) -> Bool {
...
+  let threshold = (1 - precision) * Float(pixelCount)
+  let p1: UnsafeMutablePointer<UInt8> = oldRep.bitmapData!
+  let p2: UnsafeMutablePointer<UInt8> = newRep.bitmapData!
+  for offset in 0 ..< pixelCount {
+    if p1[offset] != p2[offset]
+        || p1[offset + 1] != p2[offset + 1]
+        || p1[offset + 2] != p2[offset + 2]
+        || p1[offset + 3] != p2[offset + 3] {
+        differentPixelCount += 1
     }
```
finestructure, line 89: Mmmm, I've re-read my PR just now I think this loop is doing unnecessary extra work.

Shouldn't it advance by 4 on each loop? The four fields are R, G, B, and alpha, each 8 bit values. By the looks of it I'm comparing

0 - 0
1 - 1
2 - 2
3 - 3

and then on next iteration

1 - 1
2 - 2
3 - 3
4 - 4

instead of

5 - 5
6 - 6
...

Or maybe I'm just confused now 😅

JaapWijnen: Yeah indeed, it seems like it right? Not sure about the type and what the offset should be, but if this is true aren't you then also only checking 1/4 of all the pixels? 
So you should either calculate an offset within the loop doing:
`let actualOffset = offset * 4`
and then checking the four channels. Or just multiply the number of pixels with the amount of channels.

stephencelis: Ah huh, that makes sense. Wanna do a follow-up PR?

An alternative idea for optimization in the future would be to divide the total bytes into precision-bucketed chunks and use `memcmp` for each slice—I'm not sure this is feasible, though, and I'm not sure how often it would be faster vs. slower compared with your approach.

JaapWijnen:
It's hard to test unfortunately since photoshop likes to change the compression so too many pixels change when trying to just turn a single one red 😅

Sources/SnapshotTesting/Snapshotting/NSImage.swift
finestructure: I'll double check this asap but I think @JaapWijnen is right - this is doing overlapping checks of the first quarter of the pixels. It might be as simple as doing `offset * 4 + 0`, `offset * 4 + 1`, ... inside the loop.

I'll prep a sample image with the last pixel different to confirm.

JaapWijnen:
Found a simpler test than photoshopping haha. Suggested fix and test are up in PR #441

Sources/SnapshotTesting/Snapshotting/NSImage.swift
JaapWijnen: @finestructure sorry I was kind of intrigued so went ahead and made the PR sorry! Missed these comments.

PR #441, opened 2021-03-17
fix NSImage comparison

JaapWijnen:
Fixes an small error introduced in #435, comparing NSImages with a precision <1 was skipping some pixels. Also added a test to double check. (The test fails for the old method)


JaapWijnen:
Ah I wasn't aware XCTExpectFailure was added in the beta. Not sure how we could do this test in that case.

JaapWijnen:
removed the test as some used api was part of the new xcode beta.

JaapWijnen:
@stephencelis is this good to go like this? As diffing NSImage with precision < 1 is currently partly broken.

PR #442, opened 2021-03-21
Fix #440 - Improve WKWebView navigation delegate (option 1)

### LINKED ISSUE #440: Custom delegate on WKWebView is replaced by mock when snapshotting
When a user sets a custom navigation delegate on a WKWebView and makes a snapshot of a view containing this web view, the navigation delegate [may be replaced](https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Sources/SnapshotTesting/Common/View.swift#L631) by a stub that makes sure the snapshotting lib knows when to snapshot the web view. This works great when users don't implement a custom WKNavigationDelegate themselves. But when a user implements a custom delegate themselves this fails. 

Consider for example a custom delegate that limits the web view to github.com and opens the browser as soon as the user taps on a link outside of github.com

```
class GithubNavigationDelegate: NSObject, WKNavigationDelegate {
  override init() { }

  public func webView(
    _ webView: WKWebView,
    decidePolicyFor navigationAction: WKNavigationAction,
    decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
  ) {
    // limit to webpages on to github.com
    guard navigationAction.request.url?.host == "github.com" else {
      decisionHandler(.cancel)
      // Warn user and open url in browser instead
      return
    }

    decisionHandler(.allow)
  }
}


webview.navigationDelegate = GithubNavigationDelegate()
```

This can't be snapshot tested at the moment because the delegate is removed as soon as a snapshot is made. 

I think this could be fixed by retaining the original delegate, implementing all the possible delegate methods and forwarding all the calls to the original delegate. I'll see if I can draft up a PR. Please let me know if you think of another way of dealing with this issue. 

Finally, many thanks for all the work on this great lib 👏🏼  so far, hopefully I can contribute back a bit.

teameh:
See comments on diff + https://github.com/pointfreeco/swift-snapshot-testing/issues/440 for discussion


Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+      decidePolicy13 == aSelector,
+      original?.responds(to: aSelector) != true
+    {
+      return NSObject.responds(to: aSelector)
+    }
+
+    return interceptedSelectors.contains(aSelector)
+      || original?.responds(to: aSelector) == true
+      || NSObject.responds(to: aSelector)
+  }
```
teameh, line 35: @Sherlouk this works great. The only thing is that there are 2 optional delegate methods that might be implemented by the user:

```
func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, preferences: WKWebpagePreferences, decisionHandler: @escaping (WKNavigationActionPolicy, WKWebpagePreferences) -> Void)

func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void)
```

If the user does not choose to implement them, we should not respond to the methods, otherwise we need to. See below. 

This code is not pretty, but handeling this in the methods instead of here is even worse.

Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+    decisionHandler: @escaping (WKNavigationActionPolicy, WKWebpagePreferences) -> Void
+  ) {
+    original?.webView?(
+      webView,
+      decidePolicyFor: navigationAction,
+      preferences: preferences,
+      decisionHandler: { [weak self] policy, prefs in
+        decisionHandler(policy, prefs)
+        if policy == .cancel {
+          self?.finish()
```
teameh, line 64: If the user did choose to implement this method, we need to check if the `decisionHandler` is called with `.cancel`. If that is the case we need to call `finish` because `webView(_:didFinish)` or  webView(_:didFail:withError)` won't be called by the webView!

Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+      decisionHandler(policy)
+      if policy == .cancel {
+        self?.finish()
+      }
+    })
+  }
+
+  func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+    original?.webView?(webView, didFinish: navigation)
+    webView.evaluateJavaScript("document.readyState") { [weak self] _, _ in
```
teameh, line 85: I copy-pasted this from the original delegate. Not sure why we would execute that javascript, but then do nothing with the result. Do we execute it because webkit waits with executing it until the page is loaded?

Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+  func webViewWebContentProcessDidTerminate(_ webView: WKWebView) {
+    original?.webViewWebContentProcessDidTerminate?(webView)
+    finish()
+  }
+}
+
+fileprivate let decidePolicyPre13 = #selector(WKNavigationDelegate.webView(_:decidePolicyFor:decisionHandler:) as (
+  (WKNavigationDelegate) ->
+  (WKWebView, WKNavigationAction, @escaping (WKNavigationActionPolicy) -> Void) -> Void
+)?)
```
teameh, line 110: Also ugly, but without the type annotation the compiler can't choose between: 

```
WKNavigationDelegate.webView(_:decidePolicyFor:decisionHandler:)
WKNavigationDelegate.webView(_:decidePolicyFor:decisionHandler:)
```

And you'll get a `Ambiguous use of 'webView(_:decidePolicyFor:decisionHandler:)'`

They look the same, but actually are: 

```
original?.webView?(_: WKWebView, decidePolicyFor: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy))
    original?.webView?(_: WKWebView, decidePolicyFor: WKNavigationResponse, decisionHandler: @escaping (WKNavigationResponsePolicy))
```

¯\_(ツ)_/¯

Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+
+  if #available(iOSApplicationExtension 13.0, *, OSXApplicationExtension 10.15, *) {
+    selectors.insert(decidePolicy13)
+  }
+
+  if #available(iOSApplicationExtension 8.0, *, OSXApplicationExtension 10.11, *) {
+    selectors.insert(#selector(WKNavigationDelegate.webViewWebContentProcessDidTerminate(_:)))
+  }
+  return selectors
+}()
```
teameh, line 131: Please let me know where you want these in this file

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -1024,6 +1024,62 @@ final class SnapshotTestingTests: XCTestCase {
...
+    webView.loadHTMLString(html, baseURL: nil)
+    if !ProcessInfo.processInfo.environment.keys.contains("GITHUB_WORKFLOW") {
+      assertSnapshot(
+        matching: webView,
+        as: .image(size: .init(width: 800, height: 600)),
+        named: platform
+      )
+    }
+    _ = manipulatingWKWebViewNavigationDelegate
+  }
```
teameh, line 1051: A tests that proves that the didFinish method of the custom delegate is used

Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -1024,6 +1024,62 @@ final class SnapshotTestingTests: XCTestCase {
...
+    webView.loadHTMLString(html, baseURL: nil)
+    if !ProcessInfo.processInfo.environment.keys.contains("GITHUB_WORKFLOW") {
+      assertSnapshot(
+        matching: webView,
+        as: .image(size: .init(width: 800, height: 600)),
+        named: platform
+      )
+    }
+    _ = cancellingWKWebViewNavigationDelegate
+  }
```
teameh, line 1080: And a test that proves that the code juggling with the `webView(_ webView:decidePolicyFor:decisionHandler:)` methods also works.

Sources/SnapshotTesting/Common/ForwardingWKNavigationDelegate.swift
```diff
@@ -0,0 +1,133 @@
...
+
+#endif
```
teameh, line 133: If you want 100% coverage, we should add another test file for this class that fully tests this code without the snapshots. Please let me know, happy to provide it.

teameh:
I've also made an alternative fix: https://github.com/pointfreeco/swift-snapshot-testing/pull/443

mat1th:
Is there any update on this pr? I do face the same issue. This proposed solution does work for me.

stephencelis:
Closing in favor of #443.

PR #443, opened 2021-03-23
Fix #440 Webview loading without delegate (option 2)

### LINKED ISSUE #440: Custom delegate on WKWebView is replaced by mock when snapshotting

teameh:
While testing https://github.com/pointfreeco/swift-snapshot-testing/pull/442 another time I suddenly thought it might also be possible to remove the delegate altogether and just observe the `isLoading` property on the WKWebView. 

I've implemented that in this PR. Happy to hear your thoughts. See comments below.


Tests/SnapshotTestingTests/SnapshotTestingTests.swift
```diff
@@ -1024,6 +1024,80 @@ final class SnapshotTestingTests: XCTestCase {
...
+    _ = manipulatingWKWebViewNavigationDelegate
+  }
+
+  #if os(iOS) || os(macOS)
+  func testWebViewWithRealUrl() throws {
+    let manipulatingWKWebViewNavigationDelegate = ManipulatingWKWebViewNavigationDelegate()
+    let webView = WKWebView()
+    webView.navigationDelegate = manipulatingWKWebViewNavigationDelegate
+
+    webView.load(URLRequest(url: URL(string: "https://www.pointfree.co")!))
```
teameh, line 1059: This loads an actual URL. Is that something you want to prevent in the unit tests? I can image you want to avoid it, but it's the best real life example..

stephencelis (approved):
Great solution!

teameh:
Thanks @stephencelis !

teameh:
@stephencelis any idea when you're going to branch off the next release?

teameh:
@stephencelis? It would be great if we could use this in production

teameh:
This is part of 1.9.0 as well right? Don't see it in the release notes 🤓

teameh:
This is part of 1.9.0 as well right @stephencelis? I don't see it in the release notes 🤓

teameh:
\0/ thanks for merging 👯 😃

PR #446, opened 2021-03-26
Compare image contexts in same colorspace

dflems:
I ran some of our snapshots through a lossless image optimizer (ImageOptim) and it made a lot of them significantly smaller (removing unnecessary alpha channel, converting to monochrome or 8-bit color, etc). Unfortunately, after this conversion, image comparison failed in `swift-snapshot-testing`. I think this is because the contexts were compared against one-another using different color spaces.

This PR converts both the snapshot and the reference to the same colorspace when comparing bytes: SRGB with alpha channel. I don't know if this is the best or the most efficient but it seemed consistent enough and worked against our 1,600 snapshot tests. If anyone has some better suggestions, I'm open to hearing them (generally unfamiliar with this kind of stuff)

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -62,6 +62,11 @@ extension Snapshotting where Value == UIImage, Format == UIImage {
   }
 }
 
+// remap snapshot & reference to same colorspace
+let imageContextColorSpace = CGColorSpace(name: CGColorSpace.sRGB)
+let imageContextBitsPerComponent = 8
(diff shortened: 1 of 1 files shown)


dflems:
@stephencelis

dflems:
@stephencelis :wave:

dflems:
@stephencelis :wave: again

dflems:
@stephencelis ready to give up on this PR... any feedback?

stephencelis (approved):
Ha sorry about the delay! We have a lot of projects in-flight but this one has had the stablest API, so it's gotten a little less attention. We reviewed it this morning and it looks good to merge!

Issue #414, opened 2020-12-01
assertSnapshot does not fail if view's size is changed to zero


krzysztofpawski:
We found out that we had a failing test in our repo for two weeks but we didn't notice it because tests were passing even if they shouldn't be. The source of the not passing tests was our incorrect input data but still it's a pitty that `SnapshotTesting` didn't warn us about it.

Start with such test
```
func testExample() throws {
    isRecording = true
    let view = UIView()
    view.backgroundColor = .blue
    view.frame = CGRect(x: 0, y: 0, width: 200, height: 400)
    assertSnapshot(matching: view, as: .image)
}
```
It should record the blue rectangle.
Now change the test so we assert against empty view
```
func testExample() throws {
    let view = UIView()
    assertSnapshot(matching: view, as: .image)
}
```
And try this:
```
func testExample() throws {
    isRecording = true
    let view = UIView()
    assertSnapshot(matching: view, as: .image)
}
```
**Result:**
With second attempt test is passing.
With third attempt "empty" view is recorded.

**Expected result:**
With second attempt test is failing.

nuno-vieira:
The suggested fix is here btw: https://github.com/pointfreeco/swift-snapshot-testing/issues/414 @stephencelis feel free to provide feedback or a different solution if you think there is a better one 👍

stephencelis:
Fixed by #453.

PR #453, opened 2021-04-08
Fix false positive when asserting reference image with empty image

nuno-vieira:
Hello 👋 

### Bug Description
Unfortunately in the project that I'm working on, we encountered some false positives, where we thought that the tests were actually working, but they were not. What happened was that after some refactorings, some of the tests passed, but actually, the new reference images were the red empty image and the framework assumed that everything was okay.

In order to unit test this, I previously recorded a snapshot with content, and then verify against an empty view and check that it fails. 

Fixes: https://github.com/pointfreeco/swift-snapshot-testing/issues/454, a duplicate of https://github.com/pointfreeco/swift-snapshot-testing/issues/414


nuno-vieira:
@mbrandonw @stephencelis Any update on this, please? 🙏

Fab1n:
@mbrandonw @stephencelis Any update on this, please?

Fab1n (approved):
This looks very good in my eyes. Any concerns from the authors?

Fab1n (approved):
Now I got the review right

gabrielmoraees (approved):

stephencelis (approved):
Thanks!

Issue #447, opened 2021-03-28
Fails to build on Xcode 12.5 Beta 3


CraigSiemens:
We building either be cloning the repo or with Carthage, the following errors are produced.

```
.../swift-snapshot-testing/Sources/SnapshotTesting/AssertInlineSnapshot.swift:43:3: Cannot find 'XCTFail' in scope
.../swift-snapshot-testing/Sources/SnapshotTesting/AssertInlineSnapshot.swift:152:22: Type 'XCTContext' has no member 'runActivity'
.../swift-snapshot-testing/Sources/SnapshotTesting/AssertSnapshot.swift:52:3: Cannot find 'XCTFail' in scope
.../swift-snapshot-testing/Sources/SnapshotTesting/AssertSnapshot.swift:280:22: Type 'XCTContext' has no member 'runActivity'
```

Seems like it could be related to `libswiftXCTest` being replaced in the Xcode 12.5 releases. A similar issue occurred with https://github.com/Quick/Nimble/issues/855 where they updated to use `lXCTestSwiftSupport`.

I quickly tried adding a "Link Binary with Libraries" build phase with XCTest.framework and it successfully built. But I haven't been able to test whether it works correctly when imported into other tests.

amadeu01:
I'm still facing the same problem with Xcode 13

CraigSiemens:
You'll have better luck creating a new issue than replying on this one. The Merged PR fixed the issue for us.

PR #455, opened 2021-04-08
fix: set ENABLE_TESTING_SEARCH_PATHS to YES

thedavidharris:
Per Xcode 12.5's release notes:

```
Xcode no longer includes XCTest’s legacy Swift overlay library (libswiftXCTest.dylib). Use the library’s replacement, libXCTestSwiftSupport.dylib, instead. For frameworks that encounter build issues because of the removal of the legacy library, delete the framework and library search paths for libswiftXCTest.dylib and add the ENABLE_TESTING_SEARCH_PATHS=YES build setting. This setting automatically sets the target’s search paths so that it can locate the replacement XCTest library. (70365050)
```

### DIFF
--- a/SnapshotTesting.xcodeproj/project.pbxproj
+++ b/SnapshotTesting.xcodeproj/project.pbxproj
@@ -1016,6 +1016,7 @@
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_BITCODE = NO;
+				ENABLE_TESTING_SEARCH_PATHS = YES;
 				FRAMEWORK_SEARCH_PATHS = "$(inherited) $(PLATFORM_DIR)/Developer/Library/Frameworks";
 				INFOPLIST_FILE = Sources/Info.plist;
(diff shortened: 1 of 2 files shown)


CraigSiemens:
There's an issue I created for this a couple weeks ago https://github.com/pointfreeco/swift-snapshot-testing/issues/447.

I was just about to make this same PR. https://github.com/pointfreeco/swift-snapshot-testing/compare/main...CraigSiemens:xcode-12-5

I've tested it by adding it to out project using Carthage and it builds and runs fine. I also tried `swift build` and `pod lib lint` and both seemed to build as well.

stephencelis:
:wave: Thanks for taking the time to submit a PR! We use [XcodeGen](https://www.github.com/yonaskolb/XcodeGen) to automatically generate our Xcode project file for Carthage. Can you make the changes there instead, and then run `make xcodegen` from the root of the repo to regenerate the project file?

thedavidharris:
@stephencelis ah I missed that, done!

stephencelis (approved):
Thanks!

amadeu01:
I'm still experiencing issues with SPM and xcode 13. Hava you had any sort of issue ?

PR #467, opened 2021-05-06
Add HEIC format support

giulio92:
Apple added support for High Efficiency Image File Format (HEIC) since the following OS versions:

- iOS 11.0+
- macOS 10.13+
- Mac Catalyst 13.0+
- tvOS 11.0+
- watchOS 4.0+

the benefit of using HEIC instead of PNG is that it can store as much as image quality as PNG, but with a smaller file size.


stephencelis:
@giulio92 Thanks for the PR! It'd be great to support HEIC, as snapshots can already weigh down a repository. I think we'd want to remain backwards compatible with existing suites, though, and as implemented it seems like it would not be. While a `format` option would be one way of introducing HEIC to existing strategies, it might be a heavy solution to plumb through everywhere. Alternatively, have you considered publishing a `swift-snapshot-testing-heic-image` plugin that we could advertise from the README?

dflems:
@stephencelis @giulio92 I feel like this could probably be made backwards-compatible. New snapshots could be heic, but if there's an existing png for that test, maybe it should continue using it?

alexey1312:
Hey everyone, I made [SnapshotTestingHEIC](https://github.com/alexey1312/SnapshotTestingHEIC), maybe it will be useful for someone.

stephencelis:
Thanks @giulio92 for the PR! We're going to close it in favor of @alexey1312's plugin, which is also linked to from our README.

We'd like to make the plugin situation for the next major release of SnapshotTesting to be easier, so that great contributions like these aren't blocked so much.

PR #472, opened 2021-05-24
Use the correct scale for the diff image

codeman9:
The `UIGraphicsBeginImageContextWithOptions` was setting the `scale` parameter to `0` which will default to the device's `UIDevice.main.scale`. This results in an incorrect difference image when taking snapshots for different devices. Instead, use the maximum scale of one of the passed in `UIImage`s.


jpsim (approved):

stephencelis:
Looks good, thanks for submitting!

stephencelis (approved):

PR #477, opened 2021-06-16
Support SwiftUI snapshots on macOS with `NSImage`

MaxDesiatov:
Currently, snapshots for SwiftUI views aren't support on macOS, since there's no extension on `Snapshotting` where `Value: SwiftUI.View, Format == NSImage`. Such conformance is added, which reuses some of the code for `NSView` snapshotting.


stephencelis:
Would be nice to merge this but I'm not sure how to provide a consistent environment for that test to pass. On CI it seems to be outputting the wrong size, and locally things are failing for me, as well (visually things are identical, so might be something subtle like a color profile issue).

MaxDesiatov:
Yeah, I'm pretty sure that color profile and display scale (1x on CI and 2x Retina locally) are the culprits. I wonder if there's a way to enforce those in the tests, in the same way we can enforce dark/light mode appearance?

tillhainbach:
> I wonder if there's a way to enforce those in the tests, in the same way we can enforce dark/light mode appearance?

I did a little bit of experimenting, initially trying to use CoreImage to resample the snapshot to generic color scheme and size with no success (colors will match but the antialiasing is different). So my guess was, not matter what, one has to manipulate the local [Quartz Display Services](https://developer.apple.com/documentation/coregraphics/quartz_display_services). I found this [gist](https://gist.github.com/bellbind/27220148ea752335259dfce63f43bd5d) and [this](https://github.com/avibrazil/RDM) which reveal private apis of CoreGraphics for setting the display mode. I set my Macbook to a low-res mode, et voilà, the snapshot test passed on CI. 🎉 🥳 
Playground repo with passing test [here](https://github.com/tillhainbach/MacSnapshottesting)

PR #478, opened 2021-06-16
Introduce a way to treat snapshot recordings as artifacts

codeman9:
Our use case would prefer having the option of treating snapshot recordings (record mode as well as non-existent files) as artifacts, placing them in the same directory as failures. This PR introduces a new env option, TREAT_RECORDINGS_AS_ARTIFACTS to accomplish this goal. When set to true, new snapshot recordings will be placed in the same directory as SNAPSHOT_ARTIFACTS. When set to false, the intention is to have the same behavior as before.


codeman9:
There are some things about this implementation that I do not like so much. I will be looking into updating it a bit. One thing that doesn't quite work the same is that it overwrites the snapshot upon failure since it normalizes the file urls.

PR #481, opened 2021-06-22
Added fuzzy compare for CGImage, NSImage, and UIImage.

JWStaiert:
Includes unit and performance tests. Also fixed undercounting of components in NSImage assertSnapshot that prevented .image(precision: 0.0) from succeeding. Added package dependency to https://github.com/JWStaiert/SnapshotCompare (contains pixel comparison code.)

Pixel comparison is written in objective-c. NSImage and UIImage assertSnapshot performance versus vanilla is nearly equal under best-case scenario and more than 30 times better under worst-case (all pixels compared) on my Intel Core i7.

I will be adding fuzzy comparison for NSView, UIView, and SwiftUI Views once I understand the vanilla implementations.

Documentation has not yet been updated.


stephencelis:
@JWStaiert This is really cool! We're at the point where we'd like to simplify the core library, and are trying not to add new strategies directly to this repo. For 2.0 we are even planning on making the core library smaller but better support a plugin ecosystem. Would you be open to release your strategies as a separate library that we can link to from the README?

JWStaiert:
Yes. How can I help you achieve this?

JWStaiert:
I've added a new pull request for just those changes to SnapshotTesting needed to allow dependent packages to define snapshot strategies.

AlbertMontserrat:
Do you know when we would be able to have this Fuzzy comparison available to use it?

We are having a mix of Intel vs Silicon machines in our team, and we cannot run snapshot testing correctly without lowering down a lot the precision in some cases of big shadows. 

Comparing pixels-by-pixel with a small precision change would be really great compared to the actual approach where the precision is checking only the number of pixels different.

JWStaiert:
It's available now if you're willing to use my fork of this repo. I was asked to refactor my changes into package dependent on swift-snapshot-testing, and to do so required changes to the visibility of various types. My fuzzy compare package is at https://github.com/JWStaiert/SnapshotTestingEx and it has as a dependency my fork and development branch of this repo.

I'm looking for an open-source project to support, so if I can help you get started let me know and I'll do what I can.

Issue #429, opened 2021-02-11
Disable filename counter


h0shy:
Is there a way to disable the counter at the end of the filename? We use the iOS version in the end of the name and to have `.1` after each name is confusing. Any help would be much appreciated :)

odisei369:
You could add `named:` parameter. Then library will not add this counter at the end of a file.

Example:
```swift
assertSnapshot(matching: view, as: .image, named: "my_view")
```

I hope it helped and we could close the issue.

dafurman:
With the following example:
```
class MyViewControllerSnapshotTests: XCTest {
    func testLayout() {
        assertSnapshot(matching: MyViewController(), as: .image)
    }
}
```

Using "my_view" in a test would result in the following file name:
`MyViewControllerSnapshotTests/testLayout.my_view.png`
I think I'm in the same situation as @h0shy, where I'd really just be happy with a parameter or static var that disables the counter / suffixing, to have the full file path be: `MyViewControllerSnapshotTests/testLayout.my_view.png`.

I usually try to have one test correspond with one snapshot, so the simplest naming scheme for me would be to have the snapshot images match the test names, without adding any suffixes.

stephencelis:
Hi @dafurman! Thanks for the PR. I'm going to close this issue in favor of tracking the problem there, where we can discuss things further.

Deco354:
@stephencelis The fork containing this PR has been deleted. Can we reopen this issue?

dafurman:
Yup, +1 to the issue being reopened. I mistakenly closed that PR while cleaning out some of my Github repo forks. I forgot about that PR. I recreated it here: https://github.com/pointfreeco/swift-snapshot-testing/pull/747

PR #486, opened 2021-07-07
Add option to disable counter suffixes

dafurman:
This addresses the issue here: https://github.com/pointfreeco/swift-snapshot-testing/issues/429

This allows users to disable the default behavior of snapshot creation, where a counter is added whenever a test doesn't specify a test name.

For tests that only validate one snapshot, adding a suffix to these tests seems excessive. For me, this would result in most test files looking like `MyViewControllerSnapshotTests/testLayout.1.png` and it would be nice to be able to disable this functionality when it's not desired to keep file names simpler: `MyViewControllerSnapshotTests/testLayout.png`

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -5,6 +5,9 @@ import XCTest
 ///     diffTool = "ksdiff"
 public var diffTool: String? = nil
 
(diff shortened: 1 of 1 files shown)


dafurman:
@stephencelis Just checking in on this to see if there's any action you'd need from me here for this PR to be accepted.
Ultimately, this would help devs with a workflow like my own where there's only one snapshot per test function, with no need for suffixing being added to file names.

Deco354:
@stephencelis +1 this would be nice. I thought this was an issue on our end at first.

PR #487, opened 2021-07-07
Added snapshotDirectory as a parameter to assertSnapshot functions

dafurman:
I like using the `assertSnapshot()` functions to wrap handling of `verifySnapshot()` and asserting with the message it returns, but I'd also like the flexibility to supply a custom `snapshotDirectory` as well with these wrapping functions, just as we can do with `verifySnapshot()`.

I didn't see a clear reason why this useful parameter was left out of these functions, so I've added it in this PR.


dafurman:
Closed because there's pretty clear and detailed documentation in `verifySnapshot()` explaining how this function is designed to be wrapped and customized by users if needed.

PR #490, opened 2021-07-13
Dev snapshot compare2

JWStaiert:
Just those changes needed to allow a package dependent on SnapshotTesting to define new strategies. I'm uncertain if this is what you had in mind when you asked me to refactor my SnapshotCompare code into a plug-in. Let me know and I will revise as necessary.

Changes include: making various structs public, updating README.md to contain a link to my strategy package.


stephencelis:
@JWStaiert Sorry for missing this README addition. We've finally moved forward with a fit that addresses the core problem by merging #628, so you may be able to retire your plug-in if you don't want to maintain it. Thanks for providing this solution to others in the meantime, though!

PR #491, opened 2021-07-22
Sort URL query items

mihai8804858:
I have an issue with snapshotting an `URLRequest` that contains multiple query items. When I'm building the `URLRequest` I'm using a `[String: Any]` dictionary for queries which results in random order of query items in the final `URLRequest`.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
+++ b/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
@@ -46,12 +46,11 @@ extension Snapshotting where Value == URLRequest, Format == String {
     var components = ["curl"]
 
     // HTTP Method
-    if let httpMethod = request.httpMethod {
-      switch httpMethod {
-      case "GET": break
-      case "HEAD": components.append("--head")
-      default: components.append("--request \(httpMethod)")
-      }
+    let httpMethod = request.httpMethod!
+    switch httpMethod {
+    case "GET": break
+    case "HEAD": components.append("--head")
+    default: components.append("--request \(httpMethod)")```
     }
 
     // Headers
--- a/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
+++ b/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
@@ -76,9 +76,7 @@ extension Snapshotting where Value == URLRequest, Format == String {
     }
 
     // URL
-    if let url = request.url?.sortingQueryItems() {
-      components.append("\"\(url.absoluteString)\"")
-    }
+    components.append("\"\(request.url!.absoluteString)\"")
 
     return components.joined(separator: " \\\n\t")
   }
--- a/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
+++ b/Sources/SnapshotTesting/Snapshotting/URLRequest.swift
@@ -12,7 +12,7 @@ extension Snapshotting where Value == URLRequest, Format == String {
   /// - Parameter pretty: Attempts to pretty print the body of the request (supports JSON).
   public static func raw(pretty: Bool) -> Snapshotting {
     return SimplySnapshotting.lines.pullback { (request: URLRequest) in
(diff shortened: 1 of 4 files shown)


Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -46,17 +46,18 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
-    switch httpMethod {
-    case "GET": break
-    case "HEAD": components.append("--head")
-    default: components.append("--request \(httpMethod)")
+    if let httpMethod = request.httpMethod {
+      switch httpMethod {
+      case "GET": break
+      case "HEAD": components.append("--head")
+      default: components.append("--request \(httpMethod)")
+      }
```
mbrandonw, line 54: Let's leave these lines as the original:

```suggestion
    let httpMethod = request.httpMethod!
    switch httpMethod {
    case "GET": break
    case "HEAD": components.append("--head")
    default: components.append("--request \(httpMethod)")```
```

I'm not aware of any time that `httpMethod` can actually be `nil`, see for example:

```swift
var request = URLRequest(url: .init(string: "http://www.apple.com")!)
request.httpMethod // "GET"
request.httpMethod = nil
request.httpMethod // "GET"
```

I'm guessing this is `nil` due to an oversight during early Swift optionality audits, and so rather assume its non-`nil` until there's reproduction steps to show how it can be made `nil`.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -46,17 +46,18 @@ extension Snapshotting where Value == URLRequest, Format == String {
...
     }
 
     // Headers
     if let headers = request.allHTTPHeaderFields {
       for field in headers.keys.sorted() where field != "Cookie" {
-        let escapedValue = headers[field]!.replacingOccurrences(of: "\"", with: "\\\"")
+        guard let escapedValue = headers[field]?.replacingOccurrences(of: "\"", with: "\\\"") else { continue }
```
mbrandonw, line 60: Let's also leave the force unwrap in here, after all `field` can only be a key in the `headers` due to us looping over it:

```suggestion
        let escapedValue = headers[field]!.replacingOccurrences(of: "\"", with: "\\\"")
```

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -76,8 +77,20 @@ extension Snapshotting where Value == URLRequest, Format == String {
     }
 
     // URL
-    components.append("\"\(request.url!.absoluteString)\"")
+    if let url = request.url?.sortingQueryItems() {
+      components.append("\"\(url.absoluteString)\"")
+    }
```
mbrandonw, line 82: Force unwrapping here is fine:

```suggestion
    components.append("\"\(request.url!.absoluteString)\"")
```

mbrandonw (approved):
Looks good, thanks for taking the time to do this! 

I just have a few changes that I will take in before merging.

Sources/SnapshotTesting/Snapshotting/URLRequest.swift
```diff
@@ -50,7 +50,7 @@ extension Snapshotting where Value == URLRequest, Format == String {
     switch httpMethod {
     case "GET": break
     case "HEAD": components.append("--head")
-    default: components.append("--request \(httpMethod)")
+    default: components.append("--request \(httpMethod)")```
```
mbrandonw, line 53: Oops, accidentally left this in when commenting...

```suggestion
    default: components.append("--request \(httpMethod)")
```

mbrandonw:
@mihai8804858 Looks like the curl tests are failing because they weren't similarly sorted. I suggest either adding sorting to that snapshot strategy, or removing the curl tests.

mihai8804858:
@mbrandonw Added back query sorting for `curl`.

PR #508, opened 2021-08-21
Diff tool arguments support

arietis:
Added the `diffToolArgs` variable which allows slightly better customization of the diff message.


stephencelis:
@arietis Diff tool arguments can usually be passed before the file names, no? So couldn't you define them in the `diffTool`?

```swift
SnapshotTesting.diffTool = "vimdiff -v"
```

sunshinejr:
@stephencelis there might be commands that would want the order of the arguments in tact, e.g. I think for imagemagick you would do the images -> output image for diff at the end, so it would make sense to have the optional arguments.

stephencelis:
Closed by #840.

PR #509, opened 2021-09-01
Disabled the workaround for inaccurately rendered safe area.

grigorye:


stephencelis:
Hi @grigorye, we're not sure what to make of this PR, sorry! It's a draft without a description and we're not sure why it's disabled the safe area workaround. I'm going to close it for now but if you flesh out an explanation and motivation for the PR in the future, please do submit a new one!

Issue #510, opened 2021-09-08
Leak when running with host application


llinardos:
Here's a demo showing the leak: [SnapshotTestingLeakDemo.zip](https://github.com/pointfreeco/swift-snapshot-testing/files/7128483/SnapshotTestingLeakDemo.zip).

Go to tests, and run `testSnapshottingAViewControllerDoesntLeak`.

```
    func testSnapshottingAViewControllerDoesntLeak() {
      let vc = UIViewController()
      addTeardownBlock { [weak vc] in
        XCTAssertNil(vc, "VC should have been deallocated. It's a leak")
      }
      assertSnapshot(matching: vc, as: .image)
    }
 ```

The cause of the leak is:
When running with host application, rootViewController is retained by the window, and it retains the view controller we are testing (which is a child view controller). This causes a leak on the view controller we're testing.

The fix I propose:
By removing the view controller from it's parent when the test finishes, no one retains the view controller, fixing the leak.

NOTE:
I'm showing the leak in an external project because there's no host application in the codebase, so I can't reproduce it without adding a new test target that uses a new host application. I think is pretty noisy. But I'm open to do that.

PR #511, opened 2021-09-08
Leak when running with Host Application

llinardos:
Fixes https://github.com/pointfreeco/swift-snapshot-testing/issues/510

The cause of the leak is in the issue description 👆
The fix: By removing the view controller from its parent when the test finishes, no one retains the view controller, fixing the leak.

The same demo project that's in the issue using a branch with the fix: [SnapshotTestingLeakDemo-Fixed.zip](https://github.com/pointfreeco/swift-snapshot-testing/files/7128489/SnapshotTestingLeakDemo-Fixed.zip)

It would be great to have this automated, but it requires a new test target that uses a host application (and also a target for the host application). I'm open to doing that, but it's adding targets to the project, not sure if you want that.

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -941,6 +941,10 @@ private func add(traits: UITraitCollection, viewController: UIViewController, to
 
   return {
     rootViewController.beginAppearanceTransition(false, animated: false)
+    viewController.willMove(toParent: nil)
+    viewController.view.removeFromSuperview()
+    viewController.removeFromParent()
+    viewController.didMove(toParent: nil)
     rootViewController.endAppearanceTransition()
     window.rootViewController = nil
   }


Sources/SnapshotTesting/Common/View.swift
```diff
@@ -941,6 +941,10 @@ private func add(traits: UITraitCollection, viewController: UIViewController, to
 
   return {
     rootViewController.beginAppearanceTransition(false, animated: false)
+    viewController.willMove(toParent: nil)
+    viewController.view.removeFromSuperview()
+    viewController.removeFromParent()
+    viewController.didMove(toParent: nil)
```
stephencelis, line 947: Can you refactor your PR so this is the only change? We'd like to keep the project file and `project.yml` as is.

llinardos: ✅

stephencelis (approved):
Looks good, thanks!

Issue #530, opened 2021-10-31
Provide option to disable auto-recording on missing files


hebdavepaul0:
When using various retry scenarios on CI/CD it can be possible for a test run to auto-record a missing snapshot, and then for a subsequent run to pass on that new snapshot.

We could provide a global boolean ( similar to isRecording ) in order to enable / disable auto-record functionality.

The default value would be true so existing integrations would not be affected.

It would be up to the client test target to determine how to best toggle the boolean based on their use case.

hebdavepaul0:
implementation should be pretty straightforward, I should have a PR up by tomorrow morning at the latest.

hebdavepaul0:
PR at https://github.com/pointfreeco/swift-snapshot-testing/pull/531

stephencelis:
Thanks for the PR! I'll close this issue as a dupe so it can be tracked with the PR, instead.

PR #531, opened 2021-11-01
Allow clients to disable auto-record feature (#530)

hebdavepaul0:
PR for issue #530.

Adds a global boolean to control whether or not the framework will auto-record missing snapshots.

Helpful for cases where CI/CD might attempt multiple retries, and auto-recording snapshots could cause false successes.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -5,6 +5,9 @@ import XCTest
 ///     diffTool = "ksdiff"
 public var diffTool: String? = nil
 
+/// Whether or not to automatically record snapshots when an existing file is not found.
+public var recordMissingSnapshots = true
+
 /// Whether or not to record all new references.
 public var isRecording = false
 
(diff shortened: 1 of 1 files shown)


jklingen:
I tested the suggested change locally and for me it works as expected. 

If there is anything I can help with to get this merged (e.g. further testing), please let me know.

nunogoncalves:
We need this functionality as well! I upvote this!

I was just testing this and it works well.  I'm just thinking whether or not it makes sense to also include this as an argument to the `assertSnapshot` function, just as `record` is. 

@stephencelis @mbrandonw  do you think that makes sense? 

Also, maybe this deserves a mention on the Readme, and we would probably need a bump? Unless maybe Brandon and Stephen take care of these once more PRs get merged?

markst:
This is useful to us to be able to disable recording missing snapshots on a per architecture basis:

```swift
#if arch(x86_64)
public var recordMissingSnapshots = false
#else
public var recordMissingSnapshots = true
#endif
```

broomburgo:
Thanks for this! It would be really useful for us, for the same reason mentioned in the OP.

nunogoncalves:
@stephencelis @mbrandonw will this still be considered? by the way, @hebdavepaul0 this has conflicts. 🙏🏼

mbrandonw:
Closed via #867.

PR #532, opened 2021-11-03
Add support for Windows, add Windows job to `ci.yml`

MaxDesiatov:
This adds support for Windows, testing it with both Swift 5.4 and 5.5. I've also expanded macOS testing matrix and cleaned up one for Ubuntu to make them consistent.


stephencelis (approved):
Thanks! This looks good to us!

PR #533, opened 2021-11-06
feat(macOS): deterministically snapshot `NSView`s on any device

### LINKED ISSUE #305: macOS Screenshots with different Resolutions
When creating Screenshots on my MBP the Screenshot is captured on my retina display with high resolution and high pixel count. 

When I try to run the testing on my CI it's only rendered within the 1024x786 virtual display without retina. The Screenshots are tried to compare on complete different levels and the screenshot is failing.

What would be the best way to correct this? Do you have any hints for me to fix this in code? Would be happy to file a Pull Request about this behaviour.

### LINKED ISSUE #313: Snapshot taken on CI has slight color differences compared to local
Hi everybody, I'm having an issue here with color discrepancies in an image view. I'm using KIF framework for UI Tests and SnapshotTesting to take screenshots between interactions (asserts made on the app key window). Everything works fine, except that I get color differences from snapshots taken on my local machine vs snapshots taken on CI machine (both running the same stack, Mojave + Xcode 11.3).

Some examples below:

This one is taken by CI machine
asdf

This one is taken by my local computer
testOpenAndClose 1

This is the comparison of both images made by `imagemagick`
result

And these are the color discrepancies of a same pixel, measured  by macOS `Digital Color Meter`
WhatsApp Image 2020-03-17 at 10 11 38
WhatsApp Image 2020-03-17 at 10 11 22

I know there are lots of variables and also using KIF, but do you guys have any idea of what might be wrong?

Thanks!

### LINKED ISSUE #428: How to snapshot a view on macOS with provided screen scale?
When snapshot testing views on macOS recorded image size depends on the main display of the machine on which tests are run.

Example:
- I want to snapshot NSView of size **640x480 points**
- When running a test on a MacBook with retina display, the snapshot image will have the size of **1280x960 pixels**
- When running the same test on the same, but closed MacBook with external, non-retina screen connected, the snapshot image will have the size of **640x480 pixels**

When snapshot testing views on iOS, I can provide a display scale trait collection to force the output image scale (as described in #427). Unfortunately, I haven't found a way to snapshot test macOS views, so the test results are not dependent on the display I am using (MacBook-embedded or external monitor).

I tried to work-around this issue with the code below:

```swift
extension Snapshotting where Value == NSViewController, Format == NSImage {
  static func unscalledImage(precision: Float = 1) -> Snapshotting {
    Snapshotting<NSView, NSImage>
      .unscalledImage(precision: precision)
      .pullback { $0.view }
  }
}

extension Snapshotting where Value == NSView, Format == NSImage {
  static func unscalledImage(precision: Float = 1) -> Snapshotting {
    SimplySnapshotting<NSImage>
      .image(precision: precision)
      .pullback { $0.toImage().unscaled() }
  }
}

private extension NSView {
  func toImage() -> NSImage {
    let cacheRep = bitmapImageRepForCachingDisplay(in: bounds)!
    cacheDisplay(in: bounds, to: cacheRep)
    let image = NSImage(size: bounds.size)
    image.addRepresentation(cacheRep)
    return image
  }
}

private extension NSImage {
  func unscaled() -> NSImage {
    let image = NSImage(size: size)
    image.addRepresentation(unscaledBitmapImageRep())
    return image
  }

  func unscaledBitmapImageRep() -> NSBitmapImageRep {
    let imageRep = NSBitmapImageRep(
      bitmapDataPlanes: nil,
      pixelsWide: Int(size.width),
      pixelsHigh: Int(size.height),
      bitsPerSample: 8,
      samplesPerPixel: 4,
      hasAlpha: true,
      isPlanar: false,
      colorSpaceName: .deviceRGB,
      bytesPerRow: 0,
      bitsPerPixel: 0
    )!
    imageRep.size = size
    NSGraphicsContext.saveGraphicsState()
    NSGraphicsContext.current = NSGraphicsContext(bitmapImageRep: imageRep)
    draw(at: .zero, from: .zero, operation: .sourceOver, fraction: 1)
    NSGraphicsContext.restoreGraphicsState()
    return imageRep
  }
}
```

And while it fixes the mismatch of the output image size (with the code above it will always equal to point-size of the view), it does not work as expected. It looks like scaling the image introduces minor distortion which causes test failures (for example, text on the snapshot image is slightly shifted 1 px).

Is there a way of forcing the scale of a snapshot image on macOS?

tillhainbach:
### Summary
This PR builds upon #412 but maintains compatibility with existing API-behaviour.  Additionally to fixing `backingScaleFactor` issues, it also fixes the color space issues discussed in #477. Therefore all relevant rendering parameters can be set in a deterministic way allowing macOS snapshot test to run on any device including CI.

### In Detail

As far as I can tell from experimenting, the rendering of `NSView` is determined by its `.size`, `.appearance` `.window.backingScaleFactor`, and `window.colorSpace`. This PR tackles the later two and adds a `GenericWindow` subclass of `NSWindow` which can be configured with a specific `backingScaleFactor` and `colorSpace`. The `.image` snapshot strategy on `NSView` is extended to allow the caller to optionally set a specific `NSAppearance` and to render the view in a `GenericWindow`. This window can be configure to use CI compatible values (eg. `backingScaleFactor: 1.0` and `colorSpace: .genericRGB`). There is even a handy preconfigured static instance `GenericWindow.ci`. 

### Added Tests

Two tests are added to showcase the CI-compatibility. These are the CI-compatible version of pre-existing macOS NSView tests:

- `testNSViewCICompatible`
- `testNSViewWithLayerCICompatible`

### Caveats/Gotchas

The current implementation for `.image` for `NSView` has side-effect on the view which are not undone. This means that the calling order for `.image` and `.recursiveDescription` strategy matters. I added a note to the doc comment.

### Minor Fixes along the way

I noticed some test failures on my local machine that where:
- related to light/dark mode setting on test machine. I refactor the failing tests to explicitly use `.aqua` (light mode) appearance (the one used for provided target snapshot)
- related to number formatting using localised formatting (decimal separator in my case). I changed the number formatter to use the a `.` as the decimal separator.

### Other

This PR adds an additional step to the GitHub actions which uploads any failing snapshot artefacts. This is great for debugging failures on CI.

### Related PRs and Issues

- closes #412
- fixes issues with #477. Supporting SwiftUI, however, could be simplified to just pullback to NSView.
- closes #428
- closes #313
- closes #305

PS: I left the WIP and intermediate commits to improve comprehensibility for review but will squash them once this PR is approved to reduce the noise.


Jeehut:
What's the state of this? Is there any plan to getting this merged? Looks like there's no SwiftUI support on macOS yet. With Apple getting serious with SwiftUI on macOS now with Ventura (the new Settings app was written with SwiftUI it looks like), I'd love to see proper support for SwiftUI in this lib as well. Are there any blockers right now?

Sources/SnapshotTesting/Snapshotting/NSView.swift
```diff
@@ -9,17 +9,48 @@ extension Snapshotting where Value == NSView, Format == NSImage {
...
   ///   - size: A view size override.
-  public static func image(precision: Float = 1, perceptualPrecision: Float = 1, size: CGSize? = nil) -> Snapshotting {
+  ///   - appearance: The appearance to use when drawing the view. Pass `nil` to use the view’s existing appearance.
+  ///   - windowForDrawing: The choice of window to use when drawing the view. Pass `nil` to ignore.
+  public static func image(
+    precision: Float = 1,
+    perceptualPrecision: Float = 1,
+    size: CGSize? = nil,
+    appearance: NSAppearance? = NSAppearance(named: .aqua),
+    windowForDrawing: GenericWindow? = nil
```
stephencelis, line 27: A few questions about this.

- Any reason why this is `nil` by default and not a `GenericWindow()` that produces more deterministic snapshots? Is it just to keep compatibility with existing snaps? We're prepping a release soon that has other bug fixes that may break some existing snapshots, so we are OK making it quite clear in the release notes that upon upgrading, folks should regenerate and audit their snapshot suites. 
- Is it possible to make `GenericWindow` a private concern of the library and instead introduce options for `backingScaleFactor` and `colorSpace`? Would we be missing out on other options?
- Alternately, could this be a more generic `NSWindow` and then we could default it to be the deterministic 2x scale and RGB color space?

stephencelis (commented):
Looks really good! Thanks for PRing and sorry for the long overdue review.

I've rebased things and gotten CI building again (we want to run against a matrix of Xcode versions, so it'll just be less reliable for now).

firetop0011 (approved):

Obbut:
Are there any plans to get this merged?

For anyone finding and needing this before it is merged, I created a branch in fork which basically combines this PR as well as #477: https://github.com/Obbut/swift-snapshot-testing/tree/updated-swiftui-macos

Obbut:
If the comments from @stephencelis are blocking to get this merged, I'd be happy to create a new PR with that feedback implemented

tillhainbach:
> If the comments from @stephencelis are blocking to get this merged, I'd be happy to create a new PR with that feedback implemented

Would be great if you can take over. I was wanting to address this but I can't find the time since I haven't been working on anything swift related 🥲.

brzzdev:
@Obbut Did you ever get the time to look at this? These look like great changes

Sources/SnapshotTesting/Snapshotting/CGPath.swift
```diff
@@ -123,6 +123,7 @@ extension Snapshotting where Value == CGPath, Format == String {
 
 private let defaultNumberFormatter: NumberFormatter = {
   let numberFormatter = NumberFormatter()
+  numberFormatter.decimalSeparator = "."
```
doozMen, line 126: moved this changes into #889

Sources/SnapshotTesting/Snapshotting/NSBezierPath.swift
```diff
@@ -87,6 +87,7 @@ extension Snapshotting where Value == NSBezierPath, Format == String {
 
 private let defaultNumberFormatter: NumberFormatter = {
   let numberFormatter = NumberFormatter()
+  numberFormatter.decimalSeparator = "."
```
doozMen, line 90: moved this changes into #889

Issue #549, opened 2021-12-14
Add additional method to add XCResult attachments


mattrobmattrob:
Add an additional env variable aside from `__XCODE_BUILT_PRODUCTS_DIR_PATHS` to tell `SnapshotTesting` to write XCResult attachments. This could simplistically be just another environment variable (let's say `SNAPSHOT_TESTING_WRITE_ATTACHMENTS`) that test environments can set to write the attachments.

https://github.com/pointfreeco/swift-snapshot-testing/blob/0259c2fbd69c7bfaa34b0b86280496f84653e396/Sources/SnapshotTesting/AssertSnapshot.swift#L279-L285

This is useful in the context of Bazel where the test runner is custom but avoids setting the `__XCODE_BUILT_PRODUCTS_DIR_PATHS` env var which could cause some unintended consequences when set by something other than Xcode.

stephencelis:
Hey @mattrobmattrob! Thanks for filing this issue and opening the associated PR. For the moment I'm going to close this issue in favor of tracking it in the PR, instead.

PR #550, opened 2021-12-14
Add additional env override for XCAttachment writing

mattrobmattrob:
Add `SNAPSHOT_TESTING_WRITE_ATTACHMENTS` as an additional way to force writing the `XCTAttachment`s that end up in the xcresult bundle. This avoids having external build systems (like Bazel or Buck) use `__XCODE_BUILT_PRODUCTS_DIR_PATHS` to force writing the attachments just to be safe.

Closes https://github.com/pointfreeco/swift-snapshot-testing/issues/549.


mattrobmattrob:
Closing since this has been ignored for long enough to become out of date. 👍

PR #552, opened 2021-12-21
Added `Snapshotting.json` for `Any` value

NachoSoto:
Currently the only way to snapshot `Any` value is using `dump`. However, the format isn't always as readable as it could be.
For data that can be encoded in JSON, this new overload provides a much more readable output.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/Any.swift
+++ b/Sources/SnapshotTesting/Snapshotting/Any.swift
@@ -7,6 +7,7 @@ extension Snapshotting where Format == String {
   }
 }
 
+@available(macOS 10.13, *)
 extension Snapshotting where Format == String {
   /// A snapshot strategy for comparing any structure based on their JSON representation.
   static var json: Snapshotting {
--- a/Sources/SnapshotTesting/Snapshotting/Any.swift
+++ b/Sources/SnapshotTesting/Snapshotting/Any.swift
@@ -7,6 +7,23 @@ extension Snapshotting where Format == String {
   }
 }
 
+extension Snapshotting where Format == String {
+  /// A snapshot strategy for comparing any structure based on their JSON representation.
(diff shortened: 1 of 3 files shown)


mbrandonw:
The library ships with a `.json` strategy that works for `Encodable` types:

https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Sources/SnapshotTesting/Snapshotting/Codable.swift

And we have some tests to show how it works:

https://github.com/pointfreeco/swift-snapshot-testing/blob/6bc120bb9119f7e65b87a3c29a798b0f57f103e2/Tests/SnapshotTestingTests/SnapshotTestingTests.swift#L171-L179

Does this cover your use case?

NachoSoto:
In the test that I added I'm using an `Encodable` type for convenience, but not all types are `Encodable`.

The specific use case I'm covering here is this: I have a type from an old Swift implementation that does "manual" JSON encoding into a `[String: Any]`. I'm working on converting that to use `Codable`, but before I do I wanted to add a snapshot test for that `[String: Any]` to ensure that the new `Codable` implementation doesn't change the output.

mbrandonw:
Ah yes, I see now. Two things then: 1.) I think the strategy needs to be public, and 2.) could you update the test to use something that is not encodable, such as a `[String: Any]` dictionary?

NachoSoto:
Oh yes my bad, changed to `public`.

As for the test, it's already using `Any` (the output of `JSONSerialization.jsonObject`). That's a `[String: Any]`, but I can cast it if you prefer.

mbrandonw:
Ah ok, I breezed past that. We can leave it as-is.

NachoSoto:
Anything else I can do? :)

mbrandonw (approved):

NachoSoto:
Any chance we can get a release including this?

PR #554, opened 2021-12-29
Fixes display scaling bug

arnopoulos:
Noticed an issue where images would be saved at different scaled resolutions even when a display scale was specified (ie: `.image(traits: .init(displayScale: 3.0))`). This fixes this issue by passing along the traits parameter passed into the image function.


stephencelis (approved):
Good catch, thanks!

PR #564, opened 2022-01-24
Assign an empty image if a view of zero size is being snapshotted

davisg123:
The previous logic was "If an image has a zero frame, make the comparison image the same as the reference image"

This means that tests incorrectly pass if the reference image exists and the comparison view has a size of zero


simondelphia:
* Note I think this is the same as https://github.com/pointfreeco/swift-snapshot-testing/pull/453

stephencelis:
@simondelphia Good catch. Thanks @davisg123 for the PR! We merged the first submission, but appreciate the time you took to fix the problem.

Issue #424, opened 2021-01-28
Snapshots on Apple Silicon devices


luisrecuenco:
It seems that Apple Silicon machines generate snapshots that are slightly different from the ones generated on x86 machines. The problem is very similar to the one where the very same simulator generates different images depending on the OS version.

The issues persist when running the simulator using Rosetta.

Taking into account that it’ll be quite common in the near future to have a team of developers with Intel and Apple Silicon architectures… What’s the best way to handle it?

Thanks

ldstreet:
One possible (bad) solution could be to allow for snapshot variants based on architecture. Unfortunately, this would mean recording multiple times which would almost certainly become a nightmare.

gobetti:
this and a few other open issues seem to be duplicates of https://github.com/pointfreeco/swift-snapshot-testing/issues/313 but knowing of more scenarios leading to the same error is great for prioritizing. I've also recently upgraded to an M1 and can confirm the difference image has pixels off most usually by 1, which isn't visible in the image itself (looks plain black) but post-filtering with a simple Threshold tool will reveal those pixels

jjatie:
I have implemented a custom assertSnapshot in the referenced PR above for anyone needing to work in Silicon/Intel mixed teams. You can find it [here](https://github.com/danielgindi/Charts/pull/4574/files#diff-0d32a7d53d8fd03be1ffad9d1ec32e012e94cc23c3c299665402fec709f1832b) on line 22.

While I don't think it's a good solution, I don't know what a better alternative would be.

voidless:
In my case snapshots began rendering identically under Apple Silicon after I added 'Host Application' for unit test targets.
Currently we use iOS 14.2 simulator for unit tests, there are rendering differences with 14.4.

I had to use Rosetta because Xcode 12.4 running under Rosetta can't launch unit tests without host app.
We run Xcode under Rosetta because we currently have 9 dependencies that don't support Apple Silicon

Before that I considered adding precision:0.99 to asserts, but it's also not a good solution. In my case there were differences in rounded corners clipping

luisrecuenco:
Hey @voidless. Interesting to know about the "Host Application" fixing the issue. Unfortunately, the "Host Application" cannot be always used, so I'm afraid that's not an option for a lot of us.

voidless:
Can you give an example? We use host apps generated by cocoapods, they only have almost empty appdelegate.m, main.storyboard files and info.plist and are linked with the corresponding framework.

[...]

PR #571, opened 2022-02-09
Allow for subpixel deviations + improved unoptimized loop execution in image comparison

pimms:
## Subpixel difference

Workaround for the issue described in #313 and #424.

The readme clearly states that the reference-device should be identical to the recorder-device, but as M1 Macs are becoming popular with developers at a different rate to CI & cloud providers, that is becoming increasingly difficult.

Allow each subpixel _(R, G, B)_ to deviate by up to a certain degree. Currently, a `precision: 0.9` means that _"90% of the pixels must match **100%**"_. The issues described in particularly #313, where some experience that the pixels deviate by only 1, can be resolved by comparing with `subpixelThreshold: 1`. Comparing with `precision: 0.9, subpixelThreshold: 1` then means _"90% of the **subpixels**' value must be ±1 of the reference"_.

## Optimization

When an image comparison `memcmp` fails and we need to compare each byte of the failing image, the current loop iteration is slow when compiled without optimization on Intel machines.

My team defaulted to adding custom compiler flags to `SnapshotTesting`, but the problem can also be diminished significantly by using a more basic loop implementation.

The difference between looping over a range _(current implementation)_ and looping with a condition can be seen on Godbolt. 

<details>
<summary>Example in Godbolt</summary>

1. Go to godbolt.org
2. Change language to Swift
3. Copy the following code in the left-hand editor
4. Wait for the compiler
5. Compare sections `LBB1_4`-`LBB1_10` with `LBB3_1`
```
func byteDiff1(reference: [UInt8], comparison: [UInt8]) -> Bool {
    let maxDiffs = 100
    var diffs = 0
    for byte in 0..<reference.count {
        if reference[byte] != comparison[byte] {
            diffs += 1
            if diffs >= maxDiffs {
                return false
            }
        }
    }
    return true
}

func byteDiff2(reference: [UInt8], comparison: [UInt8]) -> Bool {
    let maxDiffs = 100
    var diffs = 0
    var byte = 0
    while byte < reference.count {
        if reference[byte] != comparison[byte] {
            diffs += 1
            if diffs >= maxDiffs {
                return false
            }
        }
        byte += 1
    }
    return true
}
```


bstien:
This is an great fix for a weird problem, and I just want to voice my support in the hope that this PR gets merged.

We have the same issue with differences in produced colors on M1 and Intel and this will solve our issue until the whole team (and possibly CircleCI) is on M1. Lowering the precision threshold is not an option for us, since we primarily use this library to test placement and size of elements. Minor differences in colors, however, is something we can excuse since we now need to have a hybrid M1/Intel solution in our team.

I've opened PRs in our component libraries that will use @pimms fork until this (hopefully) gets merged.

https://github.com/finn-no/FinniversKit/pull/1094
https://github.com/finn-no/finnui-ios/pull/66

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -99,12 +101,19 @@ private func compare(_ old: UIImage, _ new: UIImage, precision: Float) -> Bool {
...
   guard let newerData = newerContext.data else { return false }
   if memcmp(oldData, newerData, byteCount) == 0 { return true }
-  if precision >= 1 { return false }
+  if precision >= 1 && subpixelThreshold == 0 { return false }
   var differentPixelCount = 0
-  let threshold = 1 - precision
-  for byte in 0..<byteCount {
-    if oldBytes[byte] != newerBytes[byte] { differentPixelCount += 1 }
-    if Float(differentPixelCount) / Float(byteCount) > threshold { return false}
+  let threshold = Int(1.0 - precision * Float(byteCount))
```
dmitri-zganiaiko, line 106: Let's say, the image is 10x10 pixels. `byteCount` is then `10 * 10 * 4 = 400` bytes.

With `precision: 1.0` the `threshold` is then `1.0 - 1.0 * 400 = 1 - 400 = -399`.

The check `if differentPixelCount >= threshold {` will always be `true`.

It should have extra parentheses like that:

```swift
let threshold = Int((1.0 - precision) * Float(byteCount))
```

In addition, I would propose using `round` like

```swift
let threshold = Int(round((1.0 - precision) * Float(byteCount)))
```

pimms: Thanks @dmitri-zganiaiko, well spotted 👍

Resolved in 0efeef4.

fnazarios (approved):

BarredEwe (approved):

Deco354:
@dmitri-zganiaiko Nice find on the thresh-hold I think @pimms has addressed it and this is ready for review again.

dmitri-zganiaiko (approved):

Deco354:
This is still blocked I think we need an approval from someone with write access.
@stephencelis @mbrandonw are you able to help with this?

choulepoka:
Just to emphasize the important of this PR, we have a team that begins to use M1 as the main development platform, but our CI will be on Intel for the foreseeable future (At least until Github provides M1-Based Github Runners).

Sources/SnapshotTesting/Snapshotting/SwiftUIView.swift
```diff
@@ -28,11 +28,13 @@ extension Snapshotting where Value: SwiftUI.View, Format == UIImage {
   /// - Parameters:
   ///   - drawHierarchyInKeyWindow: Utilize the simulator's key window in order to render `UIAppearance` and `UIVisualEffect`s. This option requires a host application for your tests and will _not_ work for framework test targets.
   ///   - precision: The percentage of pixels that must match.
+  ///   - subpixelThreshold: The byte-value threshold at which two subpixels are considered different.
   ///   - size: A view size override.
   ///   - traits: A trait collection override.
   public static func image(
     drawHierarchyInKeyWindow: Bool = false,
     precision: Float = 1,
+    subpixelThreshold: UInt8 = 0,
```
ncipollo, line 37: Would it maybe make sense the default the threshold to a value which will avoid the M1 vs Intel issue out of the box (ex - 5)?

pimms: I haven't seen any _"one"_  value that would make sense. Sometimes, with corner-rounding for instance, the difference is `<= 2`, while with images it seems to typically be `<= 1`. With shadows and rounded corners, it may be as high as `<= 5`. I think all of these values are anecdotal though, and that the user should have a fully explicit relationship with this value. After all, this **is** a hacky workaround, and it's probably a good idea to not enable it by default.

ncipollo: YeH, makes sense, I was just hoping there was a magic value haha.

timdolenko (approved):

JProphet89 (approved):

AngCosmin (approved):

Deco354 (approved):
@pimms Thanks for sharing this I’m using it now and it works great.

I’ve a few nits that don't really matter but on the off-chance this delays the eventual merge I'm happy to do myself if you're ok with it as they appear to be pretty trivial.

* Update the `.image` entires within [Documentation/Available-Snapshot-Strategies.md](https://github.com/pointfreeco/swift-snapshot-testing/tree/main/Documentation) to include the sub-pixel property.
* SceneKit snapshot strategy support
* SpriteKit snapshot strategy support

I wasn't able to check that the internal snapshot tests for this project work, there are a fair few failing but I suspect this is due to me not knowing what simulator to use or the tests already failing.

pimms:
@Deco354 Thanks, good points. I believe that a handful of tests are failing on `main`, but I may still have broken some more.

I don't think I'll have time to look at this this week at least, so if you or someone else wants to have a look at it it'd be great — otherwise I'll get to it at some point.

Deco354:
I've one last question, what exactly does 1 byte of the `pixelDensityThreshold` equate to? I'm not very familiar with the ins and outs of Image diffing. i.e. What value would this need to be set to for all possible pixel colors to be deemed equal?

Sources/SnapshotTesting/Snapshotting/NSImage.swift
```diff
@@ -77,14 +79,19 @@ private func compare(_ old: NSImage, _ new: NSImage, precision: Float) -> Bool {
...
+  var offset = 0
+  while offset < pixelCount * 4 {
+    if p1[offset].diff(between: p2[offset]) > subpixelThreshold {
+      differentPixelCount += 1
+      if differentPixelCount > threshold {
+        return false
+      }
     }
-    if Float(differentPixelCount) > threshold { return false }
+    offset += 1
```
katzenbaer, line 94: Is there a reason this was changed from a for-loop to a while-loop? My interpretation is that in both implementations the loop should either increment `offset` or `return false`.

Using a while loop and manually incrementing the `offset` makes this less maintainable since it's easy to introduce changes in the loop body that skip incrementing `offset` and would lead to an infinite loop.

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -99,12 +101,19 @@ private func compare(_ old: UIImage, _ new: UIImage, precision: Float) -> Bool {
...
+
+  var byte = 0
+  while byte < byteCount {
+    if oldBytes[byte].diff(between: newerBytes[byte]) > subpixelThreshold {
+      differentPixelCount += 1
+      if differentPixelCount >= threshold {
+        return false
+      }
+    }
+    byte += 1
```
katzenbaer, line 116: Same comment here about replacing the for loop with a while loop.

katzenbaer:
> One more thing, what exactly does 1 byte of the `pixelDensityThreshold` equate to? I'm not very familiar with the ins and outs of Image diffing. i.e. What value would this need to be set to for all possible pixel colors to be deemed equal?

I believe it's using the `UInt8` value representation of RGB. So a value a threshold value of `255` would mean that pixels would always match regardless of their color.

stephencelis (approved):
@pimms Thanks for taking this on! We've been backed up on some of these PRs, but are hoping to merge them slowly.

I think we'll likely merge this as is unless you want to tie up some of the loose ends mentioned above, first. Once we merge we may also make a few small changes on `main` as far as API naming goes.

Let us know what you think!

Sources/SnapshotTesting/Snapshotting/NSImage.swift
katzenbaer: Nevermind, just saw your PR description with the explanation for when compiled without optimization.

katzenbaer (approved):

pimms:
Great news, @stephencelis! 🎉 I've not had as much free time to look into the follow-ups as I thought I'd have, and I know I don't have much time coming up. If you're fine with merging this as-is, that sounds great 💯

chosa91:
@pimms Can you tell us, what flags are used?
> My team defaulted to adding custom compiler flags to SnapshotTesting, but the problem can also be diminished significantly by using a more basic loop implementation.

pimms:
@chosa91 We were using CocoaPods at the time, and added the following at the end of our Podfile. I can't fully recall what Xcode-values this would translate to.

```
post_install do |installer|
    installer.pods_project.targets.each do |target|
        if target.name == "SnapshotTesting"
            target.build_configurations.each do |config|
                xcconfig_path = config.base_configuration_reference.real_path
                puts "Setting optimization flags on #{xcconfig_path}..."
                File.open(xcconfig_path, "a") {|file|
                  file.puts 'SWIFT_OPTIMIZATION_LEVEL = -O'
                  file.puts 'GCC_OPTIMIZATION_LEVEL = 3'
                }
            end
        end
    end
end
```

chosa91:
@pimms Thanks a lot!
- [SWIFT_OPTIMIZATION_LEVEL values](https://pewpewthespells.com/blog/buildsettings.html#swift_optimization_level)
- [GCC_OPTIMIZATION_LEVEL values](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW90)

simondelphia (approved):

cameroncooke:
Will this be merged soon, would be good to be able to start testing this?

CraigSiemens:
We started using this fork in out project and it's been working great so far. Setting `subpixelThreshold: 1` was enough to handle the differences in shadow rendering across different architectures for most of our failing tests.

afanaian01:
Is there a schedule when this will be merged?

reejosamuel:
> Great news, @stephencelis! 🎉 I've not had as much free time to look into the follow-ups as I thought I'd have, and I know I don't have much time coming up. If you're fine with merging this as-is, that sounds great 💯

@stephencelis any thoughts on merging this?

sschizas:
@stephencelis any plans to merge it?

Kaspik:
@mbrandonw Trying to ping someone else from this org to take a look and merge it 🙏🏼

stephencelis:
Thanks everyone in the community for their contributions towards landing this, and their patience! We've gone ahead and merged #628, which supersedes this one, so I'm going to close this out.

Thanks again @pimms for the PR, and if you believe #628 is missing anything, please let us know!

PR #579, opened 2022-03-01
Include watchOS 4.0 in @available check

aydegee:
Include watchOS in the availability check for `sortedKeys` allowing the library to compile under watchOS. 

This wouldn’t normally be necessary given that watchOS snapshots are not currently supported. 

However Xcode seems to try and compile test targets when generating SwiftUI previews and so those targeting Apple Watch weren’t rendering. 

This was making it difficult to work with SwiftUI previews inside modules containing views intended for both phone and watch platforms.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/Any.swift
+++ b/Sources/SnapshotTesting/Snapshotting/Any.swift
@@ -7,7 +7,7 @@ extension Snapshotting where Format == String {
   }
 }
 
-@available(macOS 10.13, *)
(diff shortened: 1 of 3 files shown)


dani-gavrilov:
I also faced the need for these changes. Snapshots are an iOS-only dependency and I can run tests for watchOS without any issues. But somehow the SwiftUI preview for watchOS requires these changes.

It would be nice if someone could review and merge it.

dani-gavrilov (approved):

stephencelis:
Thanks!

PR #580, opened 2022-03-04
Add optional color precision for `UIImage` diffing strategy

maxheld:
This PR introduces the extra functionality to define the color precision of each individual pixel when asserting `UIImages`. Our experience is that varying iOS version render colors slightly differently which causes a slight shift in certain color tones. This addition lets you fine tune how far off the colors are allowed to be. When asserting `UIView`/`UIViewController` snapshots with e.g. non-deterministic rendering of shadows this can also come in handy. 

The `colorPrecision` defines the leeway each individual RGBA value has before it is considered a mismatching pixel. If any of the values overstep the threshold the respective pixel is considered a mismatch and fails the test.

The `precision` parameter is still used to determine the proportion of pixels that are allowed to fail.

This has a knock-on effect for `UIView` and `UIViewController` strategies.


esemusa:
Works like a charm!

stephencelis:
@maxheld Thanks for this! We've had several versions of this kind of fix open for a bit now, but finally decided to go with #628. Please let us know if that solution does not help, and if there is something about "color precision" that works better for you. In the future, when we begin to design a 2.0, we'll be able to take into consideration the flexible needs of plugging into these diffing algorithms.

PR #585, opened 2022-03-23
Add support for tests executed repeatedly

### LINKED ISSUE #577: Running test repeatedly results in duplicate snapshot being created.
**Describe the bug**
Running a snapshot test repeatedly with xcode's new 'run test repeatedly' functionality results in additional snapshots being created for that test.

We've a wrapper around this library which I've posted below although the identifier we're passing in to the testName does not include the iteration number at the real testSnapshot callsite.

```swift
    public func testSnapshot(_ view: UIView,
                             identifier: String = #function,
                             snapshotType: SnapshotType = .image,
                             precision: CGFloat = 0.95,
                             file: StaticString = #file,
                             line: UInt = #line,
                             waiting: TimeInterval = 0,
                             isArchitectureDependent: Bool = false) {
        semaphore.wait()
        var testID = identifier
        if isArchitectureDependent && (Architecture.isArm64 || Architecture.isRosettaEmulated) {
            testID.append("-arm64")
        }
        
        switch snapshotType {
        case .image:
            assertSnapshot(matching: view,
                           as: .wait(for: waiting, on: .image(precision: Float(precision))),
                           record: isRecording,
                           file: file,
                           testName: testID,
                           line: line)
            
        case .recursiveDescription:
            assertSnapshot(matching: view,
                           as: .wait(for: waiting, on: .recursiveDescription()),
                           record: isRecording,
                           file: file,
                           testName: testID,
                           line: line)
        }
        semaphore.signal()
    }
```

**To Reproduce**
Take any snapshot test right click on it and run "test <test name> repeatedly". At a certain iteration a number matching this iteration (2 in the screenshots below) will be appended to the snapshot reference image name causing an additional snapshot to be saved and the tests to fail.
image

image

**Expected behavior**
I'd expect the test to use only one snapshot no matter how many time they're run.

**Environment**
  - swift-snapshot-testing version [e.g. 1.8.1]
  - Xcode 13.2
  - Swift 5.5
  - OS: iOS 15

krzysztofpawski:
### Motivation
`Xcode 13` introduced new parameters to `xcodebuild` that allow to run tests repeatedly ([wwdc talk](https://developer.apple.com/videos/play/wwdc2021/10296/)).
Unfortunately `SnapshotTesting` doesn't handle such situation in a correct way. 

### Proposal
We can use `XCTestObservation` in order to clean counter map between different test cases. 

### Resolved issues
Closes #577

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -173,6 +173,7 @@ public func verifySnapshot<Value, Format>(
   )
   -> String? {
 
+    CleanCounterBetweenTestCases.registerIfNeeded()
     let recording = recording || isRecording
 
     do {
@@ -327,4 +328,25 @@ func sanitizePathComponent(_ string: String) -> String {
   return string
     .replacingOccurrences(of: "\\W+", with: "-", options: .regularExpression)
     .replacingOccurrences(of: "^-|-$", with: "", options: .regularExpression)
+}
+
+// We need to clean counter between tests executions in order to support test-iterations.
+private class CleanCounterBetweenTestCases: NSObject, XCTestObservation {
+    private static var registered = false
+    private static var registerQueue = DispatchQueue(label: "co.pointfree.SnapshotTesting.testObserver")
+
+    static func registerIfNeeded() {
+      registerQueue.sync {
+        if !registered {
+          registered = true
+          XCTestObservationCenter.shared.addTestObserver(CleanCounterBetweenTestCases())
+        }
+      }
+    }
+
+    func testCaseDidFinish(_ testCase: XCTestCase) {
+      counterQueue.sync {
+        counterMap = [:]
+      }
+    }
 }


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -170,9 +169,64 @@ public func verifySnapshot<Value, Format>(
...
+
+// MARK: - Private
+
+private func assertSnapshot<Value, Format>(
+  matching value: @autoclosure () throws -> Value,
+  as snapshotting: Snapshotting<Value, Format>,
+  named name: String? = nil,
+  record recording: Bool = false,
+  timeout: TimeInterval = 5,
+  strategyIndex: UInt,
```
krzysztofpawski, line 196: I didn't want to expose this parameter in public function. That's why I created a private function and call it from public one.

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -170,9 +169,64 @@ public func verifySnapshot<Value, Format>(
...
+}
 
+private func verifySnapshot<Value, Format>(
+  matching value: @autoclosure () throws -> Value,
+  as snapshotting: Snapshotting<Value, Format>,
+  named name: String? = nil,
+  record recording: Bool = false,
+  snapshotDirectory: String? = nil,
+  timeout: TimeInterval = 5,
+  strategyIndex: UInt,
```
krzysztofpawski, line 224: The same story as with `assertSnapshot`.

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -173,6 +173,7 @@ public func verifySnapshot<Value, Format>(
   )
   -> String? {
 
+    TestObserver.registerIfNeeded()
```
maciejpiotrowski89, line 176: Maybe the name of the type or its function could reflect why this observer registers itself and why it clears the `counterMap` variable.

tahirmt:
I have two questions regarding this
1. Would we still have `SnapshotTesting` increment the counter for multiple snapshots inside one test case?
2. I'm using Quick/Nimble for this and if the test originates from the same line, i.e. a `toEventually` expectation, would it use the same file name for each execution?

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -328,3 +329,21 @@ func sanitizePathComponent(_ string: String) -> String {
...
+        registerQueue.sync {
+            if !registered {
+                registered = true
+                XCTestObservationCenter.shared.addTestObserver(TestObserver())
+            }
+        }
+    }
+
+    func testCaseDidFinish(_ testCase: XCTestCase) {
+        counterMap = [:]
```
tahirmt, line 347: I think this should use `counterQueue`

Sources/SnapshotTesting/AssertSnapshot.swift
krzysztofpawski: How about now? Does it sound better?

krzysztofpawski:
> I have two questions regarding this
> 
> 1. Would we still have `SnapshotTesting` increment the counter for multiple snapshots inside one test case? e.g. snapshot assertion from different lines but the same test case

Yes. TestObserver is called after test case is executed so the feature of counting multiple assertSnapshot calls works as expected. 

> 2. I'm using Quick/Nimble for this and if the test originates from the same line, i.e. a `toEventually` expectation, would it use the same file name for each execution?

I've never used Quick or Nimble. Would you be so kind to check with [my forked version](https://github.com/krzysztofpawski/swift-snapshot-testing/tree/support_tests_iterations) if it works with your project? I've already checked with our where we have planty of different scenarios but definitely would be cool to have confirmation also from someone else.

Sources/SnapshotTesting/AssertSnapshot.swift
krzysztofpawski: Yes, I should! Thanks for spotting this.

tahirmt:
@krzysztofpawski I tried it out and it still produces many snapshots. The way `Quick` works is that it executes the same condition with a polling until it passes.

tahirmt:
Maybe it will make sense to make it unique based on the test line. The only thing that would break it is if someone adds two assertions on the same line

krzysztofpawski:
@tahirmt could you check it one more time? I found one more case when it was producing too many reference files in our repo.
If not maybe then you could prepare a very simple project reproducing the issue mentioned by you and attach it here?

tahirmt:
@krzysztofpawski I created a demo repo to test this using your fork https://github.com/tahirmt/test-toeventually-behavior-snapshots.

krzysztofpawski:
@tahirmt I'm afraid it can't be fixed in `swift-snapshot-testing` library without some bigger refactor. But the solution I see for you is that you can use `named` argument (pass some non empty string) to omit this counting mechanism. It's not perfect but it should do the work for your matcher.

tahirmt:
> @tahirmt I'm afraid it can't be fixed in `swift-snapshot-testing` library without some bigger refactor. But the solution I see for you is that you can use `named` argument (pass some non empty string) to omit this counting mechanism. It's not perfect but it should do the work for your matcher.

That's what I was thinking too. I was thinking of many alternatives but I can't think of a better solution for now. Thanks for looking into it though.

krzysztofpawski:
@stephencelis any possibility to review this PR?

stephencelis (approved):

PR #586, opened 2022-03-24
Attach recorded snapshot

marcelofabri:
The library already attaches failed snapshots using `XCTAttachment`. 

A nice extra would be to attach in case of newly recorded snapshots as well, since some developers are already used to going to the test results screen to inspect failed snapshots - now they can do that for new ones as well.

Captura de Tela 2022-03-24 às 9 22 36 AM


stephencelis (approved):
Thanks!

Issue #588, opened 2022-03-30
Crashes if I add my view to UIWindow before recording snapshot


tahirmt:
**Describe the bug**
If I try to run a snapshot test for a view controller but by adding it into a window myself, it causes a crash

**To Reproduce**
```swift
// And/or enter code that reproduces the behavior here.
import SnapshotTesting
import XCTest

class MyViewControllerTests: XCTestCase {
    func testMyViewController() {
        let vc = UIViewController()
        let window = UIWindow()
        window.rootViewController = vc
        window.isHidden = false

        assertSnapshot(matching: window, as: .image)
    }
}
```

**Expected behavior**
The window should be recorded as any other view does

**Environment**
  - swift-snapshot-testing version 1.9.0
  - Xcode 13.2.1
  - Swift 5.5
  - OS: iOS 15

**Additional context**

This is the console messages

```
Newly-taken snapshot does not match reference.
Test Case '-[snapshottoeventuallyTests.MyViewControllerTests testMyViewController]' failed (1.089 seconds).
Test Suite 'MyViewControllerTests' failed at 2022-03-30 15:23:21.256.
	 Executed 1 test, with 1 failure (0 unexpected) in 1.089 (1.090) seconds
Test Suite 'snapshottoeventuallyTests.xctest' failed at 2022-03-30 15:23:21.257.
	 Executed 2 tests, with 1 failure (0 unexpected) in 1.107 (1.114) seconds
Test Suite 'All tests' failed at 2022-03-30 15:23:21.258.
	 Executed 2 tests, with 1 failure (0 unexpected) in 1.107 (1.119) seconds
2022-03-30 15:23:21.262036-0400 snapshottoeventually[86546:2233788] *** -[UIWindowLayer isHidden]: message sent to deallocated instance 0x6040004b1210

```

carlos-chaguendo:
in my case I have been able to solve it:

```swift
func testMyViewController() {
        let vc = UIViewController()
        let window = UIWindow()
        window.rootViewController = vc
        window.isHidden = false

        let container = UIViewController()
        container.view =  window

        assertSnapshot(matching: container, as: .image)
}
```

tahirmt:
This is no longer an issue. Verified with 1.17.5

mbrandonw:
[...]

PR #589, opened 2022-03-31
Fix crashes when recording a UIWindow

tahirmt:
Currently if we create a window and try to record a snapshot it crashes when running the full suite. In order to fix it we can avoid adding it to other windows and view controllers so I wrote a new `image` strategy explicitly for `UIWindow` to avoid this problem. 

This differs from #366 because we don't do cleanup afterwards and the window can be recorded again if there are changes

Any feedback will be appreciated.

Note: I recorded my snapshots using Xcode 13.4.1 on M1

Edit: After looking at #366 by @zenangst I realized that the changes I did could be improved so I cleaned up the code and added more tests

Closes: #588


choulepoka:
Just being a curious party here, does this allows to snapshot a UIAlertController presented modally on top of the window?

tahirmt:
> Just being a curious party here, does this allows to snapshot a UIAlertController presented modally on top of the window?

Hmm yes and no. You need to have a host application and for some reason when you capture it the UIAlertController doesn’t seem to contain any background color for the contents. The texts and buttons are all visible though.

J0s34h:
Any updates on this ?

tahirmt:
This is no longer an issue. Verified with 1.17.5

PR #593, opened 2022-04-05
Introduce a comparison strategy as a parameter

magic146:
**Problem**
Recently we discovered that sometimes Intel and M1 machines may produce slightly different snapshots. The difference is so small that it's not possible to recognize it with a human eye. But our CI finds such snapshots not equal to reference.

**Solution**
Approaching this issue we decided to change an algorithm taking into account not only how many pixels are different from each other, but also how much they are different.

Currently existing algorithm for calculating a snapshot difference:

```
var differentBytesCount = 0
allBytes.enumerate { oldByte, newByte in
    // If the bytes are not equal then difference is increased by 1. 
    if oldByte != newByte {
        differentBytesCount += 1
    }
}
```

New heuristics showed good results in our experiments:
```
var differentBytesCount = 0
allBytes.enumerate { oldByte, newByte in
    // If the bytes are not equal then difference is increased by
    // the value from [0;1] depending on how much is the difference. 
    differentBytesCount += abs(newByte - oldByte) / 255
}
```

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -88,10 +88,13 @@ public enum ComparisonStrategy {
   case equality
   /**
    The difference is calculated as the sum of normalized differences between the old and new values taken modulo.
+   Minimum difference threshold can also be specified to ignore all such bytes where it's below this threshold.
    
-   let difference = (0..<byteCount).reduce(0) { r, i in r += abs(old[i] - new[i]) / 255 }
+   let difference = (0..<byteCount).reduce(0) { r, i in r += abs(old[i] - new[i]) < minDifference ? 0 : abs(old[i] - new[i]) / 255 }
    */
-  case subtractionModule
+  case subtractionModule(minDifference: UInt8)
+  public static let subtractionModule = ComparisonStrategy.subtractionModule(minDifference: 0)
+
   public typealias ComparisonClosure = (_ oldByte: UInt8, _ newByte: UInt8) -> Float
   /**
    The difference is calculated as the sum of the values returned from the comparison closure performed for each pair of bytes.
@@ -102,8 +105,13 @@ public enum ComparisonStrategy {
     switch self {
     case .equality:
       return { ($0 == $1) ? 0 : 1 }
-    case .subtractionModule:
-      return { abs(Float($0) - Float($1)) / 255 }
+    case .subtractionModule(let minDifference):
+      return { old, new in
+        guard max(old, new) - min(old, new) < minDifference else {
+          return 0
+        }
+        return abs(Float(old) - Float(new)) / 255
+      }
     case .custom(let comparisonClosure):
       return comparisonClosure
     }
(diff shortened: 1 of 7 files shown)


Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -76,7 +78,39 @@ let imageContextColorSpace = CGColorSpace(name: CGColorSpace.sRGB)
...
+      return { ($0 == $1) ? 0 : 1 }
+    case .subtractionModule:
+      return { abs(Float($0) - Float($1)) / 255 }
+    case .custom(let comparisonClosure):
+      return comparisonClosure
+    }
+  }
+}
+
+private func compare(_ old: UIImage, _ new: UIImage, precision: Float, comparisonStrategy: ComparisonStrategy) -> Bool {
```
Sherlouk, line 113: Is there perhaps an argument to be made where the entire comparison function is more easily swappable? Not just the method in which bytes are compared.

This could allow people to provide their own comparison solutions, or threshold matching, etc., without needing to create their entire own strategy. (it's nice to still reuse all the view loading/sizing/snapshotting mechanisms that this library provides)

magic146: Hi @Sherlouk, thank you very much for your interest in our pull request!

We originally went the same route as you and added the ability to provide a custom comparison solution via a closure: https://github.com/magic146/swift-snapshot-testing/tree/allow-custom-compare-closure

But then we realized that the comparison of images in almost 100% of cases will be based on a comparison of bitmaps. And having two parameters (the precision and the method in which the bytes are compared) should be enough. In our opinion, the rest of the logic is just a framework for creating an environment for byte comparison, and it will remain the same (or even could be optimised), regardless of the comparison strategy and precision.

Having said that, I have to agree with you that at the moment creating the entire strategy can be a little expensive: while it is feasible to use a custom `Snapshotting`, it is impossible to reuse existing algorithms from the framework. However, we are proposing to make this change a separate pull request as it does not meet our current goals.

leonspok:
Hi @Sherlouk @magic146! Just wanted to add a small note regarding my recent addition to this PR. 

During the migration to Apple Silicon devices, we discovered that in some cases, we have minor colour shade differences between M1 and Intel Macs. Those changes are so subtle that we decided that it's safe for us to ignore them. We tried different comparison strategies until we found one that works best for us. It appeared to be a slight variation of the `subtractionModule`, so we extended it. Now you can pass a minimum difference threshold. Whenever the difference between bytes is smaller than this, it's assumed to be zero.

stephencelis:
@magic146 I really like the idea of this, and think it would be a nice design for whenever we have a huge breaking change, e.g. 2.0, where we can rethink how folks can extend existing strategies. In such a case we'd probably redesign the `.image` strategy to take a comparison strategy instead of individual things like `precision:` and `perceptualPrecision:`.

For now we believe we've addressed the M1/Intel discrepancy by merging #628. In the meantime I'm going to close this out, but I do think it might be worth designing such a strategy for a future version of SnapshotTesting, or if you pull out such a newly designed strategy into its own repo, please let us know!

Issue #596, opened 2022-04-08
Cannot dump full contents of NSObject subclass


tahirmt:
**Describe the bug**
For an `NSObject` subclass I cannot use `.dump` to create a snapshot to include children

**To Reproduce**
[dump-demo.zip](https://github.com/pointfreeco/swift-snapshot-testing/files/8448079/dump-demo.zip)

```swift
// And/or enter code that reproduces the behavior here.

```

**Expected behavior**
When I write an extension of my class and override the value of `renderChildren` for `AnySnapshotStringConvertible` it should render the children for my `NSObject` subclass. Instead, it seems to always use the default value of `false` and produces an empty snapshot

**Screenshots**

This is what the dump looks like for my class

<img width="859" alt="Screen Shot 2022-04-07 at 10 37 10 PM" src="https://user-images.githubusercontent.com/12048108/162352083-f5310d45-f282-4b40-aa2e-70edc0c8e728.png">

**Environment**
  - swift-snapshot-testing version : 1.9.0
  - Xcode 13.3
  - Swift 5.6
  - OS: iOS 15

**Additional context**
Add any more context about the problem here.

tahirmt:
Is there a particular reason `renderChildren` was not defaulted on? I feel like if someone wants to use `dump` they would always want the full contents of their objects printed.

PR #597, opened 2022-04-08
Add option to dump all children

tahirmt:
Adding an option to `dump` to render all children without requiring implementations for the `AnySnapshotStringConvertible` protocol default value. 

Also create several other protocols to control how children are processed for the dump. Not too happy with the naming so happy to change thaat

- `AnySnapshotStringConvertibleIncludedNodesProvider`: Adds an option to choose which nodes to include from the children. If a type implements this protocol it can provide an array of children to include in the dump
- `AnySnapshotStringConvertibleExcludedNodesProvider`: Adds an option to exclude select children from the dump. Useful when a type has unnecessary properties we want to exclude from the dump
- `AnySnapshotStringConvertibleDumpChildNodes`: When this is implemented children will be included for  `AnySnapshotStringConvertible` such as `NSObject` subclasses
- `AnySnapshotStringConvertibleIgnoreChildNodes`: Implementing types will ignore all children from the dump.

This is also a solution to #596


stephencelis:
Hi @tahirmt! Thanks for the contribution. The "dump" snapshot strategy predates our [Custom Dump](https://github.com/pointfreeco/swift-custom-dump) library, which we think is better suited to be used for a snapshot strategy these days (and in the future). Because of this, we'd rather move in that direction (by providing a plug-in that connects Snapshot Testing and Custom Dump) than make any changes to the existing strategy. Ideally we would eventually deprecate the existing strategy in favor of the plug-in. Thanks again for this, though! If you end up releasing your own plug-in, we'd be happy to list it in the README!

tahirmt:
That's a good point. I can look into that.

PR #600, opened 2022-04-29
Make optional the prevent assertion of view with 0 width and or height. 

raonivaladares:
There is a possibility of false positives results hidden by [#280](https://github.com/pointfreeco/swift-snapshot-testing/pull/280), where we started facing them in our code base.

When we update some code that unintentionally makes the view under test have a size of .zero, the tests are still passing while it should actually fail, resulting in a false positive and hiding a possible inconsistent state.

This PR introduces an extra parameter that optionally provides the possibility to force fail or skip the failure when it happens. The `default value` for this parameter is  to `skip failure` and maintain the behaviour as it is, as mentioned on  [#280](https://github.com/pointfreeco/swift-snapshot-testing/pull/280)

## Here is a very simple exemple of how to reproduce it:

 Imagine that you have a view under test already with a reference recorded.

Screenshot 2022-04-29 at 13 01 16

Them for some reason, mistake or anything it became a 0 height.
Because of the diffable image that could not be generated, but the test succeed, 
It is a false positive.

Screenshot 2022-04-29 at 12 58 29

Now with the `shouldSkipFailureOnZeroSize`, you choose how to handle it.

Screenshot 2022-04-29 at 12 58 03


stephencelis:
While the false positives are definitely annoying, we're not sure adding another config option is the way to address the problem. Previously, 0x0 images were perhaps more annoying, though, as they'd crash a test run.

The core issue is that Snapshot Testing does not make it easy to encode failure into strategies. If strategies could `throw`, we could make failing with a 0x0 image more graceful in the future. This (as well as `async` concurrency support) are things we'd like to address in a future redesign of some of the core components of the library.

We're open to alternative solutions, but we're trying not to add surface area churn to existing APIs. Let us know what you think!

raonivaladares:
> While the false positives are definitely annoying, we're not sure adding another config option is the way to address the problem. Previously, 0x0 images were perhaps more annoying, though, as they'd crash a test run.
> 
> The core issue is that Snapshot Testing does not make it easy to encode failure into strategies. If strategies could `throw`, we could make failing with a 0x0 image more graceful in the future. This (as well as `async` concurrency support) are things we'd like to address in a future redesign of some of the core components of the library.
> 
> We're open to alternative solutions, but we're trying not to add surface area churn to existing APIs. Let us know what you think!

Agree with you, also with the new version that i think closes this issue. o/
Thanks!

Issue #573, opened 2022-02-11
Precision checking algorithm is very slow


fjcaetano:
**Describe the bug**
When snapshot testing images with precision lower than 1, if the snapshots don't exactly match the reference images, the tests take much longer to run.

 We currently have 66 snapshot tests in a framework
 | Precision | Passing Tests | Failing Tests | Pixel-Imperfect Images | Total Time |
 |-|-|-|-|-|
 | 1 | 57 | 9 | 36 | 38s |
 | 0.97 | 66 | 0 | 36 | 483s |

Those 9 failing unit-tests are recording 4 images each to account for different device sizes, so those 36 images that are not pixel-perfect are responsible for a 12.7x increase in the total duration for running 66 unit tests. The recorded images are not very large, ranging from 81KB to 255KB.

**To Reproduce**
Set the precision to any value lower than 1 and modify the tested view so they're not pixel-perfect with the reference images anymore.

**Expected behavior**
Checking whether or not the snapshots are within the accepted precision should be faster. A 12.7x increase for just 36 images is a whole different order of magnitude.

**Environment**
  - swift-snapshot-testing version 1.9.0
  - Xcode 13.2.1
  - Swift 5.5
  - iOS 15.2

fjcaetano:
I believe the issue is just this for-loop, but I don't have a concrete suggestion for how to fix it. Maybe parallelize it with a multi-threaded queue?

https://github.com/pointfreeco/swift-snapshot-testing/blob/88f6e2c0afe04221fcfb1601a2ecaad83115a05f/Sources/SnapshotTesting/Snapshotting/UIImage.swift#L105-L108

gohanlon:
FYI: There's a promising open PR (#571) that aims to address three issues involving that loop, including its surprisingly long execution time (when compiled without optimizations, as is default for "Debug" builds).

Stengo:
Another option would be to move the image diff analysis from the CPU to the GPU.
The package already includes functions to create diff images (which are part of the failure output), so throwing something like the [CIAreaAverage](https://developer.apple.com/library/archive/documentation/GraphicsImaging/Reference/CoreImageFilterReference/index.html#//apple_ref/doc/filter/ci/CIAreaAverage) Core Image filter on it would allow for very quick evaluation.

[...]

PR #628, opened 2022-09-12
Perceptual image precision + 90% speed improvement

ejensen:
## Problem

The existing image matching precision strategy is not good at differentiating between a significant difference in a relatively small portion of a snapshot, and imperceivable differences in a large portion of the snapshot. For example, these snapshots below show that a 99.5% precision value allows noticeable changes (text and color) to pass the tests, while an imperceivable background color change fails the test.
| Reference | Failes | Passes |
|:--:|:--:|:--:|
|  | Imperceivable background color difference | Significant text and color changes |
| Reference | Fail |  Pass |

## Solution

This PR adds a new optional `perceptualPrecision` parameter to image snapshotting which determines how perceptually similar a pixel must be to consider it matching. This parameter complements the existing precision parameter that determines the percentage of pixels that must be considered matching in order to consider the whole image matching.

This approach is similar to https://github.com/pointfreeco/swift-snapshot-testing/pull/571 and https://github.com/pointfreeco/swift-snapshot-testing/pull/580 but uses perceptual distance rather than Euclidean distance of sRGB values. This is significant because the sRGB color space is not perceptually uniform. Pairs of colors with the same Euclidean distance can have largely different perceptual differences. 
The left and right colors of each row have the same Euclidean distance:
sRGB

This perceptual color difference calculation is performed by the [CILabDeltaE](https://cifilter.io/CILabDeltaE) CoreImage Filter which is available on macOS 10.13+, iOS 11+, and tvOS 11+. The use of CoreImage to accelerate the image diffing results in a ~90% over the existing byte-by-byte comparison.

Additionally, when macOS 11.0+, iOS 14+, tvOS 14+ is available, the [CILabDeltaE](https://cifilter.io/CILabDeltaE) filter is joined with the [CIColorThreshold](https://cifilter.io/CIColorThreshold) and [CIAreaAverage](https://cifilter.io/CIAreaAverage) filters to accelerate the whole image comparison, resulting in a ~97% speed improvement.

## Benchmarks

| Byte-Buffer Comparison | [CILabDeltaE](https://cifilter.io/CILabDeltaE) | [CILabDeltaE](https://cifilter.io/CILabDeltaE) + [CIColorThreshold](https://cifilter.io/CIColorThreshold) + [CIAreaAverage](https://cifilter.io/CIAreaAverage) |
|:--:|:--:|:--:|
| macOS 10.10+, iOS 11+, tvOS 10+ | macOS 10.13+, iOS 11+, tvOS 11+ | macOS 11.0+, iOS 14+, tvOS 14+ |
| 0.906s - baseline | 0.086s - 90% speed improvement | 0.025s - 97% speed improvement |
| baseline-macOS-10 10-compatible | macOS-10 13-compatible | macOS-11 0-compatible |

## Addressed Issues

* https://github.com/pointfreeco/swift-snapshot-testing/issues/313
  * A small `perceptualPrecsion` value (~2%) will avoid the imperceivable differences failing assertions while still failing when there are noticeable differences
* https://github.com/pointfreeco/swift-snapshot-testing/issues/424
  * The small differences in Intel/Apple Silicon hardware accelerated anti-aliasing/shadows/blur rendering are under the [2% DeltaE value which is nearly imperceivable by the human eye](http://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e)
* https://github.com/pointfreeco/swift-snapshot-testing/issues/573
  * Thanks to CoreImage acceleration, perceptual precision comparison is 90-97% faster than the exact pixel precision comparison

## Related PRs
* https://github.com/pointfreeco/swift-snapshot-testing/pull/288
* https://github.com/pointfreeco/swift-snapshot-testing/pull/490
* https://github.com/pointfreeco/swift-snapshot-testing/pull/571
* https://github.com/pointfreeco/swift-snapshot-testing/pull/580


lukeredpath:
I've just taken this for a spin in our project - so far I've converted a few of our snapshots over to this strategy and images taken in an iOS 15/iPhone 12 simulator that would fail with inperceptible differences on an iOS 16/iPhone 14 simulator are now passing.

Unfortunately due to some other packages having a dependency on this library I was not able to switch over to this fork to test this out, I had to collate the changes into a separate file in my project and rename the strategies. I've created a gist containing the file I'm using - I've not imported everything over from this PR, so far just the `UIImage`, `UIView`, `UIViewController` and `SwiftUI.View` strategies:

https://gist.github.com/lukeredpath/9abc51d9eee349c2f209cc0431c8eb6f

jlcvp:
This PR is pure gold, thank you.

I tested it on my project and all the headache and compromises we're having here with the differences between apple silicon and intel generated snapshots are gone using this new `perceptualPrecision` flag.

Looking forward to have this merged on the main branch by the reviewers

thedavidharris:
Is there any value in having `perceptualPrecision` adjustable as a global property, or is it something that should still be set per snapshot?

ejensen:
> Is there any value in having `perceptualPrecision` adjustable as a global property, or is it something that should still only be set per snapshot?

I've added a global default in tests by defining a `defaultPerceptualPrecision` global variable and adding `Snapshotting` extensions that take precedence over the ones in the `SnapshotTesting` framework:

```swift
#if arch(x86_64)
// When executing on Intel (CI machines) lower the `defaultPerceptualPrecision` to 98% which avoids failing tests
// due to imperceivable differences in anti-aliasing, shadows, and blurs between Intel and Apple Silicon Macs.
private let defaultPerceptualPrecision: Float = 0.98
#else
// The snapshots were generated on Apple Silicon Macs, so they match 100%.
private let defaultPerceptualPrecision: Float = 1.0
#endif

// Local extensions that override the default `perceptualPrecision` value with the `defaultPerceptualPrecision` global defined above.

extension Snapshotting where Value == UIView, Format == UIImage {
    /// A snapshot strategy for comparing views based on perceptual pixel equality.
    static let image = image(perceptualPrecision: defaultPerceptualPrecision)

    /// A snapshot strategy for comparing views based on perceptual pixel equality.
    ///
    /// - Parameters:
    ///   - drawHierarchyInKeyWindow: Utilize the simulator's key window in order to render `UIAppearance` and `UIVisualEffect`s. This option requires a host application for your tests and will _not_ work for framework test targets.
    ///   - perceptualPrecision: The percentage a pixel must match the source pixel to be considered a match. [98-99% mimics the precision of the human eye.](http://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e)
    ///   - traits: A trait collection override.
    static func image(
        perceptualPrecision: Float = defaultPerceptualPrecision,
        size: CGSize? = nil,
    ) -> Self {
        image(
            drawHierarchyInKeyWindow: drawHierarchyInKeyWindow,
            precision: 1,
            perceptualPrecision: perceptualPrecision,
            size: size,
            traits: traits
        )
    }
}

extension Snapshotting where Value == UIViewController, Format == UIImage {
    /// A snapshot strategy for comparing view controllers based on perceptual pixel equality.
    static let image = image(perceptualPrecision: defaultPerceptualPrecision)

    /// A snapshot strategy for comparing view controller views based on perceptual pixel equality.
    ///
    /// - Parameters:
    ///   - config: A set of device configuration settings.
    ///   - perceptualPrecision: The percentage a pixel must match the source pixel to be considered a match. [98-99% mimics the precision of the human eye.](http://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e)
    ///   - traits: A trait collection override.
    static func image(
        on config: ViewImageConfig,
        perceptualPrecision: Float = defaultPerceptualPrecision,
        size: CGSize? = nil,
    ) -> Self {
        image(
            on: config,
            precision: 1,
            perceptualPrecision: perceptualPrecision,
            size: size,
            traits: traits
        )
    }
}
```

lukeredpath:
I can confirm that this change also fixed our issue of tests not passing on M1 when using Intel generated snapshots, or vice versa. 🎉

Kaspik:
Trying this PR on our project, where devs have M1 Ultra MacBooks and CI is M1 MacMini. Some of our snapshots are still randomly failing with "ghost" failures (attaching one example). Any idea what these could be? Do we need `perceptualPrecision < 1`?

difference_3_214AC2A3-FDF0-4A2A-A4C8-C6AC7F45D91A
failure_2_214AC2A3-FDF0-4A2A-A4C8-C6AC7F45D91A
reference_1_214AC2A3-FDF0-4A2A-A4C8-C6AC7F45D91A

lukeredpath:
@Kaspik yes if you’ve got perceptual precision at 1 it won’t behave any differently. Try 0.98.

ejensen:
> Trying this PR on our project, where devs have M1 Ultra MacBooks and CI is M1 MacMini. Some of our snapshots are still randomly failing with "ghost" failures (attaching one example). Any idea what these could be? Do we need `perceptualPrecision < 1`?

@Kaspik, yes a `perceptualPrecision < 1` is needed since the images aren't exactly the same. The `perceptualPrecision` parameter is the inverse of a [Delta E value](http://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e). The following table can be used to determine the `perceptualPrecision` that suites your needs:

Perceptual Precision Value | Description
-- | --
100% | Must be exactly equal
≥ 99% | Allows differences not perceptible by human eyes
≥ 98% | Allows differences possibly perceptible through close observation
≥ 90% | Allows differences perceptible at a glance
≥ 50% | Allows differences when more similar than opposite
≥ 0% | Allows any differences

The example images you attached have a 0.3 Delta E value. So a `.image(perceptualPrecision: 0.997)` will match the images. Other images generated might have a slightly larger or smaller Delta E value, so I generally recommend a `perceptualPrecision` around 99%, but not less than 98% since perceptible differences pass the assertions with `perceptualPrecision  < 0.98`.

Kaspik:
@ejensen Thanks Eric, that makes sense. How did you get the image difference value? Asking because I have another examples where even setting `perceptualPrecision = 0.99` fails with _ghost_ image (attaching). Trying to figure out what's the difference, what's the reason, and what's the ideal precision for cases like these (have another 30-40 cases with 0.99 not working).

Difference:
difference_3_74892A27-A068-4FC6-A845-B67FBA686C30

Failure:
failure_2_74892A27-A068-4FC6-A845-B67FBA686C30

Reference:
reference_1_74892A27-A068-4FC6-A845-B67FBA686C30

ejensen:
> @ejensen Thanks Eric, that makes sense. How did you get the image difference value? Asking because I have another examples where even setting `perceptualPrecision = 0.99` fails with _ghost_ image (attaching). Trying to figure out what's the difference, what's the reason, and what's the ideal precision for cases like these (have another 30-40 cases with 0.99 not working).
> 
> I can go with 0.98, but that sounds like a bottom limit recommended by you.

@Kaspik I'm using [a new branch that reports the actual precision of the images when the tests fail](https://github.com/ejensen/swift-snapshot-testing/compare/perceptual-percision...ejensen:swift-snapshot-testing:image-diff-reporting). It reports that the precision of the second set of images you attached is 98.59%

precision-reporting

I will PR that branch if/when this PR is merged. You could attempt to use 0.985 precision for all your test cases. Or can use the branch to find a precision for each of the snapshot cases. 

98% is a good limit since it is generally difficult for an untrained eye tell the difference and display color accuracy is often inadequate to even represent the differences within that range.

Kaspik:
Ohh nice, thanks! I'll try that branch on top of this one, great job!

cooksimo:
I have tested this out and it works great for us as well 👍

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -99,12 +103,51 @@ private func compare(_ old: UIImage, _ new: UIImage, precision: Float) -> Bool {
...
+      ]
+    )
+    guard let deltaOutputImage = deltaFilter?.outputImage else { return false }
+    let extent = CGRect(x: 0, y: 0, width: oldCgImage.width, height: oldCgImage.height)
+    let thresholdOutputImage = try? ThresholdImageProcessorKernel.apply(
+      withExtent: extent,
+      inputs: [deltaOutputImage],
+      arguments: [ThresholdImageProcessorKernel.inputThresholdKey: (1 - perceptualPrecision) * 100]
+    )
+    guard let thresholdOutputImage = thresholdOutputImage else { return false }
```
stephencelis, line 122:
```suggestion
    guard
      let thresholdOutputImage = try? ThresholdImageProcessorKernel.apply(
        withExtent: extent,
        inputs: [deltaOutputImage],
        arguments: [ThresholdImageProcessorKernel.inputThresholdKey: (1 - perceptualPrecision) * 100]
      )
    else { return false }
```

ejensen: Thanks for adding support for older Swift versions!

stephencelis (approved):
@ejensen Sorry for the delay! This looks like a great update to us. We're going to merge things and fast-track a release soon. There's a lot upstream, including some bug fixes that will require folks to rerecord some snapshots. This breakage has been holding up release, but we don't think snapshot backwards compatibility is feasible at this time, and it shouldn't hold up all the good things that have been contributed.

Do you think you can get a PR up for the two improvements you commented about?

- Adding a global configuration option for projects that expect to run tests on both Intel/M1
- Adding the extra messaging about the actual precision difference on failure

simondelphia:
If I use `0.98` for `precision` and I don't specify `perceptualPrecision` (using 1.10.0) I have no failures on my repo (as expected). But if I use `0.995` for `precision` and `0.98` for `perceptualPrecision` I get a number of failures where there's no perceptible difference. And the errors I get are saying that the `precision` is under `0.98` even though in the latter case I had the same snapshots passing with `0.98` `precision`. 

Does using `perceptualPrecision` change how `precision` is calculated?

stephencelis:
@simondelphia I believe perceptual precision will accommodate many pixels being different by a very small amount, while precision will accommodate a few pixels being completely different. Depending on your situation, you may prefer one over the other.

When it comes to generating snapshots on the exact same iOS simulator, but sometimes from M1 machines and sometimes from Intel, the "shape" of the rendered output should be exactly the same, but all of the pixels may be very slightly off.

Meanwhile, if you are running snapshots generated against iOS 15 in an iOS 16 simulator, but from the same device, many of the pixels may be identical, but font rendering between releases may cause a few pixels to be completely off.

simondelphia:
yea, just wondering why the error message I'm getting when I use `0.995` for `precision` is telling me I have `precision` below `0.98` when I use a `perceptualPrecision` below 1 alongside `precision`. But when I don't specify `perceptualPrecision` the same snapshots pass.

ejensen:
> yea, just wondering why the error message I'm getting when I use `0.995` for `precision` is telling me I have `precision` below `0.98` when I use a `perceptualPrecision` below `1.0` alongside the `precision` above. But when I don't specify `perceptualPrecision` the same snapshots pass with `0.98` `precision`.

There is a difference in how the `precision` % is calculated depending on whether perceptual precision is enabled. The difference is whether the precision is the  % of pixels matching or the % of bytes matching (each pixel has 4 bytes - one for each channel: Red, Green, Blue, and Alpha). 
* When `perceptualPrecision < 1` the new algorithm calculates % of pixels matching
* When `perceptualPrecision = 1` the old algorithm calculates the % of bytes matching

When using the old algorithm, a pixel with multiple channels different (ex Red + Green) skews the image precision more than if only one channel of the pixel were different. This seems intuitive at first but has some unintended side effects. For example, we'd expect these 2 images (solid white and solid black) to have a 0% precision:
referenceactual
However, when using the old algorithm the precision is 25%. This is because the alpha of an image is the same across the two images. Meaning 25% of the bytes match. Therefore any images with the same alpha have a floor of 25% precision when using the old algorithm. Not ideal.

The new precision algorithm that's enabled when `perceptualPrecision < 1` is simply a percentage of pixels that meet the `perceptualPrecision` requirement. Therefore it's not skewed by pixel channels, and doesn't suffer the same side effects:

Screen Shot 2022-09-23 at 12 20 29 PM

The old algorithm was left as-is `perceptualPrecision = 1` to prevent existing break tests that haven't opted-in to the perceptual precision algorithm.

simondelphia:
Hm interesting

simondelphia:
I used `precision` = `0.9` and `perceptualPrecision` = `0.98` and now all but two snapshots passed.

Here's one set:

[Archive.zip](https://github.com/pointfreeco/swift-snapshot-testing/files/9637405/Archive.zip)

> Actual image precision 0.03083378 is less than required 0.9 Actual perceptual precision 0.9574219 is less than required 0.98 (0.00s)

I'm confused by the 0.03 vs 0.9 and why is the perceptual precision almost 0.95 when there's no noticeable difference? Have a light mode version of the same snapshots but only the dark mode ones failed.

The failing image was captured on an Intel Mac on remote CI (AWS) and the reference was captured on an M1 mac locally.

The other failing set was also a dark mode version for a somewhat similar screen. 

Both have a speckled pattern in the background, which I suspect is part of the reason why it's saying there's a difference.

ejensen:
@simondelphia  When comparing the PNG images included in the zip they match with `precision = 1.0` and `perceptualPrecision = 0.99`. You should be able to replicate this by writing a test that compares the `UIImage`s similar to [`testImagePrecision()`](https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Tests/SnapshotTestingTests/SnapshotTestingTests.swift#L292-L308). I'm interested to see if the `UIImage` comparison test passes on your CI machines. If so, it means there are differences in the `UIView/UIViewController` snapshots that are lost when the snapshots are saved to PNGs. This could be CI color profile differences (possibly similar to https://github.com/pointfreeco/swift-snapshot-testing/issues/313#issuecomment-1258535742) that are normalized when saved to PNGs. If your project is adjusting the output PNGs (similar to https://github.com/pointfreeco/swift-snapshot-testing/pull/446) the color space of the PNGs could differ from the original snapshots.

I'm working on [a Swift Playground](https://github.com/pointfreeco/swift-snapshot-testing/compare/main...Frameio:swift-snapshot-testing:image-comparison-playground) that assists in debugging image differences by outputting all the intermediate values and providing suggestions for precision parameter values. It might help identify the causes of different machine rendering and suggest values that accommodate them.
NotMatching

simondelphia:
> In my test target I've added a global default by defining a `defaultPerceptualPrecision` global variable and adding `Snapshotting` extensions that take precedence over the ones in the `SnapshotTesting` framework:

@ejensen: According to https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1256439724, doesn't using a perceptual precision of 1 on Apple Silicon Macs cause the library to use the old snapshotting algorithm, and thus defeat the purpose of using `perceptualPrecision`?

simondelphia:
I tried `assertSnapshot` with the reference image above on the failed snapshot on my local M1 machine: 

| `precision`  | `perceptualPrecision` |  result | message
| ------------- | ------------- | ------------- | ------------- |
| 1 | unspecified | failure | Newly-taken snapshot does not match reference |
| 0.99 | unspecified | success | |
| 0.99 | 0.99 | success | |
| 1 | 0.99 | failure | Actual perceptual precision 0.9862305 is less than required 0.99 |
| unspecified | 0.99 | failure | Actual perceptual precision 0.9862305 is less than required 0.99 |

On CI with the same calls I got the same failures with the same precisions.

On CI the snapshots on the original login UIView used 0.98 and 0.99 for precision and perceptual, respectively:

> Actual image precision 0.02803135 is less than required 0.98 Actual perceptual precision 0.9574219 is less than required 0.99 (0.00s)

simondelphia:
@ejensen

simondelphia:
I managed to produce a diff using a tool online that gave me something I could actually see as a difference in png files from the same snapshots. Seems to be something about the edge around the capsule button, and I suppose while small there are a substantial number of pixels involved. 



Notably the two failing snapshots are also only dark mode tests and the corresponding light mode ones worked fine.

ejensen:
> According to https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1256439724, doesn't using a perceptual precision of 1 on Apple Silicon Macs cause the library to use the old snapshotting algorithm, and thus wouldn't this defeat the purpose of using `perceptualPrecision`?

On the team I'm using that strategy, everyone locally has the same M1 machine so the snapshots exactly match. We generate new snapshots locally, so everyone's local machine produces snapshots that are byte identical. There's no need for any `precision` or `perceptualPrecision` that isn't 100%. However, our CI machines are Intel, so we set the 
`perceptualPrecision < 1` when comparing snapshots in CI test runs. 
This precision-setting strategy works for this team, but if your team is generating snapshots on a mix of different machines, then you'd want to have `perceptualPrecision < 1` for all architectures.

> On CI with the same calls I got the same failures with the same messages.
> 
> On CI the snapshots on the original login UIView used 0.98 and 0.99 for precision and perceptual, respectively:
> 
> > Actual image precision 0.02803135 is less than required 0.98 Actual perceptual precision 0.9574219 is less than required 0.99 (0.00s)

This points to the CI machine producing a snapshot of the original login UIView that differs from the image that is saved. This difference might be a color space difference. This PR https://github.com/pointfreeco/swift-snapshot-testing/pull/665 normalizes the image color spaces before comparison, which might be the solution to resolve your CI image differences.

IlyaPuchkaTW:
We've started to use this new precision parameter recently to mitigate M1 vs Intel issue and have exactly the same setup - 1.0 precision for local runs on M1 and 0.999 for CI runs (we are using Bitrise). For whatever reason though while locally tests fail as expected the same tests pass on CI with such precision. At the same time when precision is 1.0 on CI they as expected fail. We have parallelised tests enabled.
Any tip what could be the reason would be welcome.

ejensen:
> We've started to use this new precision parameter recently to mitigate M1 vs Intel issue and have exactly the same setup - 1.0 precision for local runs on M1 and 0.999 for CI runs (we are using Bitrise). For whatever reason though while locally tests fail as expected (both with 1.0 and 0.999 precision, calculated precision is actually negative) the same tests pass on CI with such precision. At the same time when precision is 1.0 on CI they as expected fail. We have parallelised tests enabled.
> Any tip what could be the reason would be welcome.

The issue sounds similar to the one addressed in https://github.com/pointfreeco/swift-snapshot-testing/pull/666 where some virtualized macOS environments silently fail due to the lack of Metal support.

IlyaPuchkaTW:
Thanks, indeed CI is running on VM, I'n trying the package from your fork with those changes but it seems like it drastically affects overall performance of the tests.

tdrhq:
@IlyaPuchkaTW An alternative approach, is to use an open source tool like Vizzy or Screenshotbot, so that your screenshots are always recorded in CI (hopefully all CI machines have an identical environment).

https://github.com/workday/vizzy

https://github.com/screenshotbot/screenshotbot-oss

(I built Screenshotbot, so I'm a bit biased toward that. I know of teams using it with swift-snapshot-testing)

Issue #625, opened 2022-09-02
`WKWebView` iOS test crashing on Xcode 14 beta 6


marcosgriselli:
**Describe the bug**
iOS `testWebView` test is crashing on Xcode 14 beta 6 with the error: 
**SnapshotTesting/View.swift:759: Fatal error: Unexpectedly found nil while unwrapping an Optional value**

**To Reproduce**
Run the iOS tests on the `main` branch on Xcode 14 beta 6.

**Expected behavior**
Tests to pass

**Screenshots**

<img width="1024" alt="image" src="https://user-images.githubusercontent.com/14804033/188243779-2bfaa948-b521-4122-b9f4-b36809fa454a.png">

**Environment**
  - swift-snapshot-testing version: latest main
  - Xcode 14 beta 6
  - Swift 5.7
  - OS: 12.5.1

barrault01:
Still happening with Xcode 14 RC.

teameh:
Still happening with Xcode 14.0.1.

A crash inside WKWebView occurs that seems to make the `wkWebView.takeSnapshot(...)` method to return an error instead of an image. 

```
xctest[42043:34329227] [Process] 0x110064100 - [PID=0] WebProcessProxy::didFinishLaunching: Invalid connection identifier (web process failed to launch)
xctest[42043:34329227] [Process] 0x110064100 - [PID=0] WebProcessProxy::processDidTerminateOrFailedToLaunch: reason=Crash
xctest[42043:34329227] [Process] 0x13a82da18 - [pageProxyID=6, webPageID=7, PID=0] WebPageProxy::processDidTerminate: (pid 0), reason=Crash
xctest[42043:34329227] [Loading] 0x13a82da18 - [pageProxyID=6, webPageID=7, PID=0] WebPageProxy::dispatchProcessDidTerminate: reason=Crash
xctest[42043:34329227] [Loading] 0x13a82da18 - [pageProxyID=6, webPageID=7, PID=0] WebPageProxy::dispatchProcessDidTerminate: Not eagerly reloading the view because it is not currently visible
```

I don't think this is caused by SnapshotTesting. Just using the `takeSnapshot` method without SnapshotTesting also fails. Weird that I can't find any other blog posts or issues about this.

teameh:
@barrault01 I've opened a PR with your commit because it does improve things a little bit.

mbrandonw:
I believe #692 fixed this, so closing for now.

PR #629, opened 2022-09-15
Fixing wkWebView.takeSnapshot with Xcode 14

barrault01:
Today the lib is crashing with Xcode 14 and snapshoting WKWebviews.

This will fix this issue: https://github.com/pointfreeco/swift-snapshot-testing/issues/625

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -755,7 +755,11 @@ extension View {
                 callback(Image())
                 return
               }
-              wkWebView.takeSnapshot(with: nil) { image, _ in
+              let configuaration = WKSnapshotConfiguration()
+              if #available(iOS 13, *) {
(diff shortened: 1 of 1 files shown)


barrault01:
I will close this because it did not solve 100% the problem. It did not crash anymore but the webview content is not pictured.

teameh:
> It did not crash anymore but the webview content is not pictured.

This seems to does the trick though for iOS **15** running in Xcode 14. Could you maybe reopen the PR?

teameh:
Weird. If I try this outside of swift-snapshot-testing, this works fine, but this patch does not work for the lib..

teameh:
But applying this patch to the lib does help a little bit. Without this patch, no snapshots are made by swift-snapshot-testing and the the tests can't complete. With this patch at least we get a snapshot of the background color of the WebView.

PR #638, opened 2022-09-21
Include failure reason when snapshot comparisons fail

ejensen:
A follow-up to https://github.com/pointfreeco/swift-snapshot-testing/pull/62 specifically this comment https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1250404163

This PR updates the image comparison methods to report a failure message rather than just a boolean of whether the comparison passed or failed. This failure message provides more insight into the reason why a snapshot assertion failed. This is especially helpful when the image differences are subtle - such as the "ghost" differences mentioned in https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1250224235

Fancy logo
Fancy logo

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/NSImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/NSImage.swift
@@ -1,5 +1,4 @@
 #if os(macOS)
-import CoreImage.CIFilterBuiltins
 import Cocoa
 import XCTest
 
@@ -18,14 +17,17 @@ extension Diffing where Value == NSImage {
       toData: { NSImagePNGRepresentation($0)! },
       fromData: { NSImage(data: $0)! }
     ) { old, new in
-      guard !compare(old, new, precision: precision, perceptualPrecision: perceptualPrecision) else { return nil }
+      guard let message = compare(old, new, precision: precision, perceptualPrecision: perceptualPrecision) else { return nil }
       let difference = SnapshotTesting.diff(old, new)
-      let message = new.size == old.size
-        ? "Newly-taken snapshot does not match reference."
-        : "Newly-taken snapshot@\(new.size) does not match reference@\(old.size)."
+      let oldAttachment = XCTAttachment(image: old)
+      oldAttachment.name = "reference"
+      let newAttachment = XCTAttachment(image: new)
+      newAttachment.name = "failure"
+      let differenceAttachment = XCTAttachment(image: difference)
+      differenceAttachment.name = "difference"
       return (
         message,
-        [XCTAttachment(image: old), XCTAttachment(image: new), XCTAttachment(image: difference)]
+        [oldAttachment, newAttachment, differenceAttachment]
       )
     }
   }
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Much better!

Sources/SnapshotTesting/Snapshotting/NSImage.swift
```diff
@@ -57,61 +59,46 @@ private func NSImagePNGRepresentation(_ image: NSImage) -> Data? {
...
-      toBitmap: &averagePixel,
-      rowBytes: MemoryLayout<Float>.size,
-      bounds: CGRect(x: 0, y: 0, width: 1, height: 1),
-      format: .Rf,
-      colorSpace: nil
+    return perceptuallyCompare(
+      CIImage(cgImage: oldCgImage),
+      CIImage(cgImage: newCgImage),
+      pixelPrecision: precision,
+      perceptualPrecision: perceptualPrecision
```
ejensen, line 100: The CoreImage operations were [extracted into a separate method](https://github.com/pointfreeco/swift-snapshot-testing/pull/638/files#diff-3e1681b0623d05afda1993aa39ee2ae8641918d2e1df61acf80b1a367dada06fR174-R218) since they were common across `UIImage.swift` and `NSImage.swift`

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -189,6 +171,51 @@ private func diff(_ old: UIImage, _ new: UIImage) -> UIImage {
...
+  var maximumDeltaE: Float = 0
+  context.render(
+    deltaOutputImage.applyingFilter("CIAreaMaximum", parameters: [kCIInputExtentKey: new.extent]),
+    toBitmap: &maximumDeltaE,
+    rowBytes: MemoryLayout<Float>.size,
+    bounds: CGRect(x: 0, y: 0, width: 1, height: 1),
+    format: .Rf,
+    colorSpace: nil
+  )
+  let actualPerceptualPrecision = 1 - maximumDeltaE / 100
```
stephencelis, line 208: @ejensen I'm not familiar with this filter's output, but when taking this for a spin in isowords I noticed that it seems to be flipped:

image

Should this just be `maximumDeltaE / 100`?

stephencelis: Or maybe not, I was misremembering for this one:

IMG_3696

Might just need to get a feel for the change and what the numbers "mean" :)

ejensen: @stephencelis Delta E should have an inverse relationship with precision. A higher Delta(Difference)E means the pixels are less similar(precise).

The wording of the error reporting might need some finessing. Seeing  "Actual perceptual precision: 0.56" insinuates the entire image is significantly different. When in reality the perceptual precision being reported is the lowest of all the pixels compared, could be a significant outlier.

simondelphia: So when it says "Actual image precision 0.03083378 is less than required 0.9" what does that mean?

ejensen: > So when it says "Actual image precision 0.03083378 is less than required 0.9" in my test logs what does that mean?

That means that the lowest perceptual precision of all the pixels is 3.08%. You would need to set `perceptualPrecision <= 0.0308` if you were to require all the pixels to meet the perceptual requirement  (eg. setting `precision: 1.0`).

The failure messages could both better explain the values and provide suggestions. I'm prototyping this in the [image-comparison-playground branch](https://github.com/pointfreeco/swift-snapshot-testing/compare/main...Frameio:swift-snapshot-testing:image-comparison-playground). It contains an iOS/macOS compatible Swift Playground where you supply two snapshots and it displays the result of the comparison along with all the intermediaries for inspection.

NotMatchingMatching

The output makes suggestions for values to use when images do not match:
```
❌ Images do not match…
The images would match if either:
  1. precision is lowered to 0.9991
  2. perceptualPrecision is lowered to 0.9939
  3. precision is raised to 1.0 and perceptualPrecision is lowered to 0.9912
ℹ️ 99.91% of pixels match with 99.98% perceptual precision
```

Matching images also get suggestions when the precision values could be raised:
```
✅ Images match
The images would still match if:
  1. precision is raised to 0.9998
  2. perceptualPrecision is raised to 0.9939
  3. precision is raised to 1.0 and perceptualPrecision is set to 0.9859
ℹ️ 100.00% of pixels match with 98.00% perceptual precision
```

There is a tradeoff between detail + precision and speed of execution. It is particularly expensive to calculate the maximum matching `perceptualPrecision` when `precision < 1`. It takes a couple of seconds to calculate with a full-screen snapshot. The other suggestions are quite fast and are likely reasonable to include in the swift-snapshot-test's failure messages.

simondelphia: > Actual image precision 0.03083378 is less than required 0.9

Isn't that referring to `precision`, not `perceptualPrecision`?

simondelphia: So there is at least one pixel in the image that is 97% different from the same pixel on the reference imag?

ejensen: > So does it mean there is at least one pixel in the image that is 97% different from the same pixel on the reference image

Yes, that is what it means.

simondelphia: I guess currently there's nothing built-in that I can use to filter out outliers?

PR #641, opened 2022-09-22
Image precision default value variable

ejensen:
## Overview
A follow-up on https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1246768428

This PR adds global variables that control the default image `precision` and `perceptualPrecision` parameter values when no value is specified in the `Snapshotting.image()` method.

This makes it simple to configure snapshot tests with precision values that accommodate subtle differences between device rendering. Such as the [imperceptible differences in Intel and Apple Silicon hardware accelerated anti-aliasing, shadow, and blur rendering](https://github.com/pointfreeco/swift-snapshot-testing/issues/424).

### Alternatives considered:
Integrators could add [overloading extensions that override `Snapshotting.image()` parameter defaults](https://github.com/pointfreeco/swift-snapshot-testing/pull/628#issuecomment-1246800844). But, this requires more code to be written by the integrator and is more susceptible to breakage and infinite recursion.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -16,6 +16,30 @@ public var record: Bool {
   set { isRecording = newValue }
 }
 
+/// The default precision used in image snapshot comparisons.
+///
+/// Represents a percentage of pixels in the snapshot that must match the reference image.
+///
+/// - Precondition: Value is a percentage that should be within the range `0...1`.
+public var defaultImagePrecision: Float = 1
+
+/// The default perceptual precision used in image snapshot comparisons.
+///
+/// Represents how perceptually similar pixels must be to consider them matching.
+/// The value is the inverse of a [Delta E](http://zschuessler.github.io/DeltaE/learn#toc-defining-delta-e).
+///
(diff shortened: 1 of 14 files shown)


stephencelis:
I think this makes sense, though as you note in the alternative, nothing prevents folks from defining these overrides.

Kaspik:
We have overrides and I think it makes more sense. Global variables can be confusing/misleading sometimes.

I would love to see the second PR with the difference reporting for failures! 🙏🏼

ejensen:
> I would love to see the second PR with the difference reporting for failures! 🙏🏼

The [second PR](https://github.com/pointfreeco/swift-snapshot-testing/pull/638) was merged and included in the 1.10.0 release.

Kaspik:
Bruh, I'm slow. Thanks!

stephencelis:
@ejensen Let's keep this open as a draft to gather feedback, but if others agree that a workaround is better than more global config surface area, we can close it out.

In the future, we'd love to encode more data into snapshot formats by default, and then strategies could use this to do a "best guess" for diffing. We probably won't have time to explore such a thing anytime soon, though.

PR #668, opened 2022-10-26
Add in diffToolSuffix to support ImageMagick cli

tisohjung:
Added diffMessageConverter to support other tools like ImageMagick
which has commands that looks like `magick compare "${image1}" "${image2}" diffImage.png`

```
    diffMessageConverter = { "magick compare \"\($0)\" \"\($1)\" diff.png ; open diff.png" }
```


mlostekk:
any idea when this will be merged?

adozenlines (approved):

mkaya93 (approved):

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -7,6 +7,7 @@ import XCTest
 /// diffTool = "ksdiff"
 /// ```
 public var diffTool: String? = nil
+public var diffToolSuffix: String = ""
```
chosa91, line 10: What about rework these two into a reusable, tool-independent message converter?
I'm suggesting this due to the following.  Although, the solution offered works fine with these 2 tools, but what about using separate switches to prefix each item (for example: `another_img_diff_tool -d 42 -i img1.png -i img2.png`? 

Something like this (at the moment I have not bothered to deprecate the old variables, but that would complete the picture):

```swift
public var diffMessageConverter: (String, String) → String? = { "\(diffTool) \($0) \($1)" }
```

tisohjung: good suggestion
[8959584](https://github.com/pointfreeco/swift-snapshot-testing/pull/668/commits/8959584fcfedfd0ca93f45509e809656e564a760)

stephencelis:
Closed by #840.

PR #675, opened 2022-11-12
Make the image paths in Xcode XCTest output clickable

RufusMall:
Simple change to print absolute paths if you are not using a diffing tool. This preserves the scheme of the URL , making the experience a bit nicer as you can click on the blue text to navigate to the file in finder. Or you can right click the link and open in a tool of your choice

What I tested:
I don't have KSDIFF installed, so I just eyeballed to double check the output matches. I tested with Xcode on macOS. I don't have access to be able to test other platforms :(.

Output before - when not using a diffTool:
before_notool
Output after - when not using a diffTool (notice you can click on the links to navigate to the files in finder):
after_notool

Output before - when using KSDiff:
before_ksdiff
Output after - when using KSDiff (expected to match the above)
after_ksdiff


stephencelis (approved):
This change looks good to us, thanks!

Issue #683, opened 2022-12-08
Not handled files


valeriyvan:
**Describe the bug**
`swift build` produces warning about 227 not handles files:
```
warning: 'swift-snapshot-testing': found 227 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraits.ipad-pro-10inch-50-split-landscape.png
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testUIViewControllerLifeCycle.2.png
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraits.ipad-9-7.txt
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Fixtures__/pointfree.html
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraits.ipad-9-7.png
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraitsWithViewController.label-accessibility-extra-large.txt
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testNamedAssertion.named.txt
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraitsEmbeddedInTabNavigation.iphone-8.png
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraitsEmbeddedInTabNavigation.ipad-pro-11.png
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testAnySnapshotStringConvertible.substring.txt
/********/swift-snapshot-testing/swift-snapshot-testing/Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testTraits.ipad-pro-10-5-alternative.png
[...]

PR #684, opened 2022-12-08
Exclude snapshots and fixtures in manifest

valeriyvan:
Solves warning #683 about not handled files.

Not sure if excluding is correct way of handling. 
May be these files should be copied as resources:
```
resources: [
  .copy("__Snapshots__"),
  .copy("__Fixtures__")
]
```


stephencelis (approved):
Thanks!

PR #695, opened 2023-01-18
Add PreviewSnapshots to "Plug-ins"

jflan-dd:
We've recently released a small library for sharing configurations between SwiftUI Previews and snapshot tests


stephencelis:
Looks like a great plug-in! Thanks for sharing!

PR #715, opened 2023-02-23
Update Readme.MD

WDI-Json:
~~I've added a code snippet for using SwiftUI and also~~ extended the instructions for setting up. 

In previous instructions, it seemed unclear to me where you needed to make changes concerning the build target. Therefore, I've made the instructions more clear and added some sample code when using SwiftUI in combination with SnapshotTesting. Would love to hear your thoughts on this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Once [installed](#installation), _no additional configuration is required_. You 
 ``` swift
 import SnapshotTesting
 import XCTest
+@testable import MyTestProject
 
 class MyViewControllerTests: XCTestCase {
   func testMyViewController() {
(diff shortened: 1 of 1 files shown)


stephencelis:
Hey @WDI-Json, thanks for taking the time to submit a PR! I think the existing instructions _should_ automatically link the library to the test target, though Xcode can make it easy to get this wrong.

At the end of the day we need to strike a balance between instructions that are library-specific and general to Xcode, and I think we want to lean toward not maintaining too much general-purpose Xcode instructions, so I'm inclined to leave it as-is. What do you think?

WDI-Json:
Hey @stephencelis, 

First of all, thank you for taking the time considering my pull request. I'm still kinda new to iOS-Development and wasn't sure if this would be a good improvement to your documentation. I wrote it thinking of other junior devs just starting out with snapshots, thinking it might improve setting up.

It's a minor change nonetheless which could help out. Then again I totally get your viewpoint. 

Ps: love what you guys do with TCA!!

PR #716, opened 2023-02-28
Configurable SnapshotDirectory to support Xcode cloud workflow

rcarver:
Following the thread in #553 it seems there are two steps to using snapshots on Xcode cloud:

1. Snapshot files must be found in the `ci_scripts` directory, which is the only set of files available during a test run. This can be done via symlink.
2. `assertSnapshot` must be called with a `snapshotDirectory` argument pointing to the location of the file's snapshots within the `ci_scripts`  directory.

The project I'm working with has many swift packages, many of which use snapshots. I wanted a way to run the snapshot tests on Xcode cloud without changing the local workflow. 

This change allows configuration of the snapshot directory via ENV var, with an option specifically for this workflow.

---

## 1. Symlink snapshots

This script finds all `__Snapshot__` directories and symlinks them into `ci_scripts/snapshots`. Run it from your project root.

```bash
#!/bin/sh
set -e
snapshots=ci_scripts/snapshots
search_path_from_snapshots=../..
rm -rf $snapshots
mkdir $snapshots
cd $snapshots
find $search_path_from_snapshots -type d -name "__Snapshots__" | grep -v .build | while read dir; do
    parent=$(dirname "$dir")
    parent_name=${parent##*/}
    echo $parent_name $dir
    ln -s "$dir" "$parent_name"
done
```

## 2. Configure the snapshots location for CI

Set an ENV var to change where the library finds snapshots:

```
SNAPSHOTTESTING_PACKAGE_PATH=/Volumes/workspace/respository/ci_scripts/snapshots
```

> Note - I'm currently setting this in a Test Plan, but I think it can be set in a Scheme. Ideally in the Xcode cloud Workflow setup but those vars don't seem to be available to the test run?

The result is that snapshots from all packages are flattened and symlinked into the `ci_scripts` dir, then `SnapshotTesting` constructs a matching path for assertion when run via CI.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -119,6 +119,118 @@ public func assertSnapshots<Value, Format>(
   }
 }
 
+/// A set of options for how to save snapshots for each test.
+public enum SnapshotDirectory {
+  /// Standard directory to save snapshots.
+  ///
+  /// By default snapshots will be saved in a directory with the same name as the test file, and that directory will sit inside a directory `__Snapshots__` that sits next to your test file.
+  ///
+  ///  `.snapshotsForFile`
+  ///
+  ///   /MyProject/MyTests.swift
+  ///   /MyProject/__Snapshots__/MyTests/snapshot.1.json
+  case snapshotsForFile
+
+  /// Alternate directory to save snapshots.
+  ///
+  ///  `.path("/MyProject/configured/path")`
+  ///
+  ///   /MyProject/MyTests.swift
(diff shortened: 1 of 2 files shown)


jdowd7:
@rcarver used your fork, tried to implement per above... script works great, and my build is as expected. 
But when I run tests I get the below:

image

jdowd7:
@rcarver thanks! I'll give this a try.
Drop in meaning manually import and not using SPM?

rcarver:
@jdowd7 yes use via SPM. It defines a new `assertSnapshots` function so you should just be able to swap `import SnapshotTesting` to `import SnapshotTestingEnvOverlay`

PR #718, opened 2023-03-05
feat: added new devices (14, 14 Pro, 14 Pro, 14 ProMax)

hariom-palkar:
This PR adds the 14, iPhone14Plus,  iPhone14Pro, and iPhone14ProMax devices for snapshot testing.

I took the numbers for the device config from: https://useyourloaf.com/blog/iphone-14-screen-sizes/

I believe the traits should be the same as iPhone 13

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -241,6 +241,74 @@ public struct ViewImageConfig {
     return .init(safeArea: safeArea, size: size, traits: .iPhone13ProMax(orientation))
   }
 
+  public static let iPhone14 = ViewImageConfig.iPhone14(.portrait)
+
+  public static func iPhone14(_ orientation: Orientation) -> ViewImageConfig {
+    let safeArea: UIEdgeInsets
+    let size: CGSize
+    switch orientation {
+    case .landscape:
+      safeArea = .init(top: 0, left: 47, bottom: 21, right: 47)
+      size = .init(width: 844, height: 390)
+    case .portrait:
+      safeArea = .init(top: 47, left: 0, bottom: 34, right: 0)
+      size = .init(width: 390, height: 844)
+    }
+
+    return .init(safeArea: safeArea, size: size, traits: .iPhone14(orientation))
+  }
+
+  public static let iPhone14Plus = ViewImageConfig.iPhone14Plus(.portrait)
+
+  public static func iPhone14Plus(_ orientation: Orientation) -> ViewImageConfig {
+    let safeArea: UIEdgeInsets
+    let size: CGSize
+    switch orientation {
+    case .landscape:
+      safeArea = .init(top: 0, left: 47, bottom: 21, right: 47)
+      size = .init(width: 926, height: 428)
+    case .portrait:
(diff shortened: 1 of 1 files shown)


HenSquared:
It would be great to get this merged to get the new devices! Any idea of what is holding this up? 🤔

jlcvp (approved):

hariom-palkar:
Can we get this reviewed or merged?

thompsondave:
Can we get an update on when this change will be merged in and if the plan is to continue to actively maintain this library as there hasn't been any merge to master for about 6 months now.

tinder-kaijing:
Can we get this merged in order to support iPhone 14?

tinder-owenthomas:
@stephencelis I see you have merged several other PRs in this repo. Is there a process this PR needs to go through to get merged? Our team is working on updating to the latest devices now and would love to go to iPhone 14 if possible. Thanks!

iosdevben:
Is this library a dead project?

tinder-owenthomas:
@mbrandonw / @stephencelis is Point-Free still maintaining this repo?

mbrandonw:
Hi @tinder-owenthomas, yes we do maintain this repo, and merge pull requests that we think are important. However, this project, more so than any of our others, has parts that need constant updating due to Apple's release schedule, and it's just not feasible to keep up-to-date with it all the time. 

For example, even in this PR we think there are potentially some mistakes, such as having [forceTouchCapability](https://github.com/pointfreeco/swift-snapshot-testing/pull/718/files#diff-ac6a6bcd60f9a7be58cebc4ed15f0f1ed4d0a5264cd98add63aae7be88cc7114R830) set to `.available` even though it seems that 3D touch hasn't been supported on Apple devices for a [very long time](https://en.wikipedia.org/wiki/Force_Touch#Products). And it's probably just copy-pasta from other parts of the library, but keeping track of all of these nitty-gritty details in Apple's ecosystem doesn't really seem like a good responsibility for this library.

Further, we feel that a lot of these little utilities can be implemented outside the library, and probably should be. For example, the contents of the PR can be copy-pasted into any project. It does not require any internals from the library. The fact that we haven't yet merged this PR should not hold anyone up from bringing it into their own projects. And the same is true of almost all of the snapshot strategies that are very difficult to maintain in the face of Apple's ever changing frameworks and bugs.

In the end we'd like to evolve this library into a state where nearly all of the strategies are moved out into separate projects, largely supported by the community, so that all that is left is the core of the snapshotting tools.

rfodge:
So you won't be adding the support for the iPhone 14 devices?

mbrandonw:
Hi @rfodge, we are open to merging it, but there are some potential problems with it currently, such as auditing the usage of `forceTouchCapability`, and making sure all of the other properties are correct.

But I want to re-iterate that the library does not _need_ to support these devices. The `ViewImageConfig` type is a public struct with a public initializer. It is just a few fields for describing a few aspects of the view. I think it would be perfectly reasonable for you to define `static let`s for the few device configurations you support in your project, and to not depend on us keeping a 100% accurate list of devices in this library.

rfodge:
Ok I will give that a try thank you for the quick response

bystritskiy:
Half a year has passed :(

Issue #264, opened 2019-10-28
How do i snapshot test a scroll view?


bitwit:
I'm trying to write some snapshot tests around a UITableView and was attempting to scroll to snapshot the bottom of the table but no matter what I do the snapshot appears to ignore scoll position.

Originally i was snapshotting a view controller, but then tried snapshotting the tableview directly and got a fatal error warning me `View not renderable to image at size ...`

Am I taking the wrong approach to this or is there some way to snapshot a scrollable view to capture all its contents?

Thanks for the help. This library is a thing of beauty and I'm excited to finally be playing with it at work.

bitwit:
I think I have a solution so i will close this issue.

In case anyone else has the problem, I found that rendering to a custom image size configuration allowed me to get everything in a single image:

```swift
assertSnapshot(matching: vc, as: .image(on: .init(safeArea: .zero, size: .init(width: 480, height: 2000), traits: .init())))
```

stephencelis:
Hi @bitwit! Yeah that's the solution we typically go with. Not quite sure where it might fit to document for others, but if you have any ideas please let us know!

UrsKahmann:
Hey, did you guys by any chance find a way to take a screenshot of scroll view, which is only as large as the content? The approach that @bitwit used (with setting a fixed size) works, but requires a good guess at how heigh the scroll view will be.

bitwit:
@UrsKahmann last time I checked there is no way to only capture a page's worth of content at a time. Typically I set the height a reasonable amount higher than I anticipate so there's some wiggle room, but it's not perfect.

bitwit:
@robinsalehjan yes, this was functional on bitrise CI always. Not sure why you're getting a dimension difference if you set it manually

PR #723, opened 2023-03-16
Support UIViewController contentScrollView

lechuckcaptain:
I added an optional argument to the `.image` snapshot strategy that try to retrieve the `UIViewController.contentScrollView` (when it's available), in order to better size the target `UIViewController` in the case a `UIScrollView` is used and the content is greater than the selected device frame. 

This proposal should provide a better way of working around the issue https://github.com/pointfreeco/swift-snapshot-testing/issues/264.


dhrushit-RIT:
Is there any update to this issue?

maross:
> Is there any update to this issue?

+1

Sources/SnapshotTesting/Common/View.swift
```diff
@@ -934,6 +935,15 @@ func prepareView(
   }
   let dispose = add(traits: traits, viewController: viewController, to: window)
 
+  if #available(iOS 15.0, *) {
+      if computeScrollSize, let scrollView = viewController.contentScrollView(for: .top) {
```
chosa91, line 939: In some cases this is misleading, for example, `computeScrollSize: true` on the caller site, but without effect when the project's deployment target is less than iOS 15.

Having an alternative for backward compatibility that looks for the top-most or bottom-most(?) `scrollView` would be great.

lechuckcaptain: It makes sense. Just to better understand, can you expand a little about what do you think should be the alternative for devices with target < iOS 15 ? Thanks!

chosa91: I'm thinking about something that traverses through the subviews and returns the first (topmost) one which matches the predicate of `view is UIScrollView`.

https://stackoverflow.com/questions/32301336/swift-recursively-cycle-through-all-subviews-to-find-a-specific-class-and-appen
https://stackoverflow.com/questions/32151637/swift-get-all-subviews-of-a-specific-type-and-add-to-an-array/45297466

I'm not saying that these implementations are correct, so please use them with a grain of salt, but the idea is something like this.

PR #724, opened 2023-04-20
[Feat] globla var artifactsUrl

Wei18:
### Summary:
Would like to upload artifacts when CI running but test results failed, but `ProcessInfo.processInfo.environment["SNAPSHOT_ARTIFACTS"]` is hardly to control.

### Detail:
Like [`snapshotDirectory`](https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Sources/SnapshotTesting/AssertSnapshot.swift#L168), but artifactDirectory.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -8,6 +8,12 @@ public var diffTool: String? = nil
 /// Whether or not to record all new references.
 public var isRecording = false
 
+/// If assert failure will save artifacts in here, you can use it to uplaod artifacts when running CI.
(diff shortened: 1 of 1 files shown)


Wei18:
```
import Foundation

// Get the current environment variables
var environment = ProcessInfo.processInfo.environment
print("Current Environment: \(environment)")

// Set a new environment variable
let newEnvironmentValue = "NewValue"
setenv("NEW_VARIABLE", newEnvironmentValue, 1)  // The third parameter is used to specify whether to replace an existing variable with the same name

// Retrieve the environment variables again to check if the setting was successful
environment = ProcessInfo.processInfo.environment
print("Updated Environment: \(environment)")
```

PR #727, opened 2023-04-29
Provide a Snapshotting strategy using `UIView.systemLayoutSizeFitting`

oconnelltoby:
This allows constraining a view in one axis, whilst allowing the other to resize. This is particularly comment when snapshotting views with text in as to support accessibility text should wrap when it cannot fit horizontally.


AllDmeat:
iOS 17 changed how traits are applied: trait overrides are applied only after view is already added to the view hierarchy.
You are requesting view size before it's being added to the hierarchy. Do you have an updated implementation to support Dynamic Type testing on iOS 17?

PR #737, opened 2023-06-06
Support for xcode 15b1

finestructure:
This makes the project build on macOS 13 Ventura with Xcode 15.0b1. Tests also pass (same OS version) with Xcode 14.3.1, which seems a bit odd since there were snapshot diffs. I didn't go back and test `main` with Xcode 14.3.1 but I believe it would mean `main` wouldn't pass with Xcode 13.4.1 on macOS 13 Ventura 🤷‍♂️

Not sure if this is useful at this stage but since we use snapshot testing in SPI-Server and it failed the build due to these errors, I thought I'd start the discussion :)


Sources/SnapshotTesting/Snapshotting/NSBezierPath.swift
```diff
@@ -38,6 +40,7 @@ extension Snapshotting where Value == NSBezierPath, Format == String {
   /// A snapshot strategy for comparing bezier paths based on pixel equality.
   ///
   /// - Parameter numberFormatter: The number formatter used for formatting points.
+  @available(macOS 11.0, *)
```
finestructure, line 43: These two availability attributes are the only ones truly required to fix the build with Xcode 15.

Tests/SnapshotTestingTests/__Snapshots__/SnapshotTestingTests/testImagePrecision.exact.png
finestructure, line 1: This is a baffling diff, not sure why that comes out so differently - nor how that passes with Xcode 13.4.1 after the change.

stephencelis (approved):
Thanks! Some of the artifacts are puzzling but will investigate outside of merging.

PR #759, opened 2023-08-20
Prevent writing artifacts to invalid location

smidf:
If `SNAPSHOT_ARTIFACTS` value is dynamically obtained from an environment variable, an empty string is set for the non-existing variable. This change prevents writing the artifacts to an invalid / prohibited location.


smidf:
@stephencelis, @mbrandonw Any feedback on this small fix, guys?

PR #768, opened 2023-09-12
New `canGenerateNewSnapshots` to avoid automatically creating missing ones in CI

### LINKED ISSUE #748: [Request] Avoid automatically recording snapshots on CI
I'm experimenting with moving to Xcode Cloud, and noticed that for some reason it's not finding the snapshots even though they're in the repo.
Because we have test retries enabled, I almost didn't see this, because the second try for every test passed:
Screenshot 2023-07-20 at 13 02 36

It occurs to me that it should be able to disable this behavior when running tests on CI, because the subsequent test run will always be a pass. If there is no snapshot when running tests on CI, it shouldn't try to create them.

Bonus points if there's an environment variable that the library can detect to automatically enable that setting in Xcode Cloud / CircleCI, etc.

NachoSoto:
Fixes #748.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -11,6 +11,10 @@ public var diffTool: String? = nil
 /// Whether or not to record all new references.
 public var isRecording = false
 
+/// Whether or not to create snapshots if they aren't found.
+/// It's recommended to set this to `false` in CI to avoid false negatives if the test is retried.
+public var canGenerateNewSnapshots = true
+
 /// Whether or not to record all new references.
 ///
 /// Due to a name clash in Xcode 12, this has been renamed to `isRecording`.
@@ -256,7 +260,7 @@ public func verifySnapshot<Value, Format>(
       return "Couldn't snapshot value"
     }
 
-    guard !recording, fileManager.fileExists(atPath: snapshotFileUrl.path) else {
+    guard !recording, canGenerateNewSnapshots, fileManager.fileExists(atPath: snapshotFileUrl.path) else {
       try snapshotting.diffing.toData(diffable).write(to: snapshotFileUrl)
       #if !os(Linux) && !os(Windows)
         if ProcessInfo.processInfo.environment.keys.contains("__XCODE_BUILT_PRODUCTS_DIR_PATHS") {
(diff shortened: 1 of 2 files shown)


muh-bilal (approved):

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -11,6 +11,10 @@ public var diffTool: String? = nil
 /// Whether or not to record all new references.
 public var isRecording = false
 
+/// Whether or not to create snapshots if they aren't found.
+/// It's recommended to set this to `false` in CI to avoid false negatives if the test is retried.
```
NachoSoto, line 15:
```suggestion
/// It's recommended to set this to `false` in CI to avoid false positives if the test is retried.
```

NachoSoto:
Bump>?

CraigSiemens:
I had a related idea and stumbled across this searching issues for it.

I was looking for a way to record only the snapshots for the failing tests to avoid re-recording all the snapshots in a project. Mainly to avoid unrelated changed being applied and the back and forth of intel/apple silicon shaping out images.

---

This is just an idea, feel free to take or leave it. We could introduce a `recordingMode` rather than having more bools in the tests/globally.

```swift
enum RecordingMode {
    /// Only record a snapshot when there isn't an existing one. The default behaviour.
    case missingSnapshots

    /// Record the snapshots only for failing tests. Useful for when a change affects many
    /// snapshots across a project but you don't want the snapshots for the passing tests
    /// to be recorded as well.
    case failingTests

	/// Always record snapshots. The same as setting `isRecording` to true.
    case always

    /// Never record a snapshot. The same as setting `canGenerateNewSnapshots` to true.
    case never
}
```

This could be introduced without being a breaking change by modifying `isRecording` to be

```swift
public var isRecording: Bool {
   get { recordingMode == .always }
   set { recordingMode = newValue ? .always : .missingSnapshot }
}
```

mbrandonw:
Hi @NachoSoto, thanks for the PR!

It's taken some time, but this is now implemented in #867. I'm going to merge this PR and then it will be updated to mold into what we have going on in #867.

mbrandonw:
Also, @CraigSiemens, #867 adds your option for recording failed snapshots only (see 0c6b44916f169bf670463133432b7c301da65573). Thanks for the suggestion!

PR #772, opened 2023-09-13
Escape carriage returns explicitly

### LINKED ISSUE #771: Inline snapshot testing output fails when URL request body expected contains carriage returns
**Describe the bug**
I'm converting our assertion helper that tests our API client routes generate the correct URL requests over to the new inline snapshot tool. It works well except for one test that uses my `swift-url-routing-multipart` extension to create a multipart request body. This intentionally contains `\r\n` line breaks, but these get converted to `\n` when written into the source file using the `.raw` strategy, causing the test to fail when re-run.

```swift
func testCarriageReturnInlineSnapshot() {
    let string = "This is a line\r\nAnd this is a line\r\n"
    var request = URLRequest(url: URL(string: "https://www.example.com")!)
    request.httpMethod = "POST"
    request.httpBody = string.data(using: .utf8)!
    assertInlineSnapshot(of: request, as: .raw) {
        """
        POST https://www.example.com

        This is a line
        And this is a line

        """
    }
}
```

**Expected behavior**
I'd expect the original line breaks to be preserved.

**Environment**
  - swift-snapshot-testing 1.13.0
  - Xcode 14.3
  - OS: iOS 16

stephencelis:
Either Xcode or SwiftSyntax is stripping the carriage return out of the snapshot, but either way we should probably explicitly escape them for visibility.

Fixes #771.

### DIFF
--- a/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
+++ b/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
@@ -387,7 +387,11 @@ private final class SnapshotRewriter: SyntaxRewriter {
             segments: [
               .stringSegment(
                 StringSegmentSyntax(
-                  content: .stringSegment(snapshot.actual.indenting(with: leadingIndent))
+                  content: .stringSegment(
+                    snapshot.actual
+                      .replacingOccurrences(of: "\r", with: #"\\#(delimiter)r"#)
+                      .indenting(with: leadingIndent)
+                  )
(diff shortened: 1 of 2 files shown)


lukeredpath:
Can confirm this fixes the issue for my specific test failure. 🎉

stephencelis:
@lukeredpath Great! Will merge when green and cut a patch release in the next few days after the dust settles a bit more.

PR #773, opened 2023-09-13
doc) fix CONTRIBUTING.md

soo941226:
Hi. I'm not sure that this document is necessary. I saw your some trending repositories don't have this document.
However, this URL get to me 404 page anyway. so I fixed it.


stephencelis (approved):
Thanks!

PR #776, opened 2023-09-16
Added an isOpaque parameter to view strategies to allow for transparent snapshots

HarshilShah:
This PR allows you to capture transparent screenshots. Previously, all transparent bits would get replaced with white, and so this could be a breaking change for existing tests, and so is gated behind an `isOpaque` parameter that is true by default, thus retaining the old behaviour

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -964,6 +964,7 @@
     func snapshotView(
       config: ViewImageConfig,
       drawHierarchyInKeyWindow: Bool,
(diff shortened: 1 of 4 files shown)


adozenlines (approved):

HarshilShah:
Update: I've just updated this PR to explicitly set the format of the generated image to generate PNGs without alpha channels when the image is opaque. This helped me work around an issue where App Store Connect wasn't accepting screenshots that had alpha channels even though they had no transparent pixels and were rendered with `isOpaque: true` or with the `main` branch of this project

PR #785, opened 2023-10-05
Update .spi.yml

finestructure:
Fix watchOS builds by selecting correct scheme

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -11,4 +11,4 @@ builder:
       - SnapshotTesting
       - InlineSnapshotTesting
     - platform: watchOS
-      scheme: SnapshotTesting
+      scheme: InlineSnapshotTesting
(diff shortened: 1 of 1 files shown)


stephencelis:
@finestructure Just to understand, does this mean watchOS isn't building InlineSnapshotTesting?

finestructure:
It wouldn't, no. We only support a single scheme and we've got to pick one here. Unless the repo came with an xcodeproj file that sets up a different scheme that builds both of the other schemes (or targets, in SwiftPM parlance).

It really comes down to which target/scheme you prefer as a compatibility indicator.

Actually, I've just looked at the package manifest and `InlineSnapshotTesting` depends on `SnapshotTesting`. So the easiest thing is just to use `InlineSnapshotTesting` as the "driving" scheme. I'll update the PR.

stephencelis (approved):
Thanks!

PR #795, opened 2023-10-13
Widen swift-syntax dependency to 508.0.1..<510.0.0

stephencelis:
Following discussion from #794, we can be a better citizen in the ecosystem and play nicely with those that have stricter dependencies.

I've made `assertInlineSnapshot` simply unavailable when another dependency locks to 508.0.1. I think there are too many significant changes in swift-syntax 509 to support the function in earlier versions.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
     ),
   ],
   dependencies: [
-    .package(url: "https://github.com/apple/swift-syntax.git", from: "509.0.0")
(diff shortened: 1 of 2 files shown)


mbrandonw (approved):
👍 Seems reasonable! 

/cc @groue

groue:
This is a complete success, thank you very much! This solves #794 completely. :-)

I'll wait a little bit more for inline snapshot testing 😉

PR #809, opened 2023-11-14
Add record mode to inline snapshots

stephencelis:
Makes it a little more convenient to record individual snapshots over and over.


Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
```diff
@@ -114,7 +116,7 @@ import Foundation
         )
         return
       }
-      guard let difference = snapshotting.diffing.diff(actual, expected)?.0
+      guard let difference = snapshotting.diffing.diff(expected, actual)?.0
```
stephencelis, line 119: This diff felt backwards in the error message, so this should improve the readability.

mbrandonw (approved):

PR #811, opened 2023-11-19
Exclude newlines from indent string in SnapshotRewriter.

scogeo:
This PR fixes an issue that was first raised in pointfreeco/swift-macro-testing#11. I mysteriously hit this same issue of extra newlines in the generated expansion while editing a macro test source file that was previously working. It was a real head scratcher, but I was able to track it down.

The underlying cause is the algorithm to determine the default indentation level in `SnapshotRewriter`.  The code looks for the first use of leading whitespace in the source file, and then uses that to determine the indention size.  However, it didn't account for the case where the first use of leading whitespace, may be an empty "indented line" followed by a newline, with no non-whitespace characters.

A trivial example would be:

```swift
import Foundation
import InlineSnapshotTesting
import SnapshotTesting
import XCTest

final class WhitespaceTests: XCTestCase {
  
  func testInlineSnapshot() {
    assertInlineSnapshot(of: ["Hello", "World"], as: .dump)
  }
}
```

In the above example there are two spaces after the class declaration and before the function declaration which causes the indention string to be calculated as `"  \n"` instead of the expected `"  "`, resulting in extra newlines being introduced in the generated output.  When run the following is produced:

```swift
import Foundation
import InlineSnapshotTesting
import SnapshotTesting
import XCTest

final class WhitespaceTests: XCTestCase {
  
  func testInlineSnapshot() {
    assertInlineSnapshot(of: ["Hello", "World"], as: .dump) {
      
"""

▿ 2 elements

  - "Hello"

  - "World"


"""
    }
  }
}
```

With this PR, the correct output is now produced:

```swift
import Foundation
import InlineSnapshotTesting
import SnapshotTesting
import XCTest

final class WhitespaceTests: XCTestCase {
  
  func testInlineSnapshot() {
    assertInlineSnapshot(of: ["Hello", "World"], as: .dump) {
      """
      ▿ 2 elements
        - "Hello"
        - "World"

      """
    }
  }
}
```

### DIFF
--- a/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
+++ b/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
@@ -389,7 +389,7 @@ public struct InlineSnapshotSyntaxDescriptor: Hashable {
       self.indent = String(
         sourceLocationConverter.sourceLines
           .first(where: { $0.first?.isWhitespace == true && $0 != "\n" })?
-          .prefix(while: { $0.isWhitespace })
+          .prefix(while: { $0.isWhitespace && $0 != "\n" })
           ?? "    "
       )
       self.snapshots = snapshots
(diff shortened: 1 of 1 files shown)


adozenlines (changes requested):
.prefix(while: { $0.isWhitespace && !$0.isNewline})

scogeo:
@stephencelis I think this may have gone overlooked over the holidays. I would have thought it would have been merged by now. Perhaps my original PR comment wasn't clear enough so I've included a trivial example you can copy and paste into a file to verify the issue. Make sure to preserve all of the whitespace!

scogeo:
I noticed this was recently fixed in #830, so I'm closing this PR.

PR #812, opened 2023-11-27
Improve the speed of comparing memory buffers by using a workaround t…

mbrandonw:
…o a missed compiler optimization

Rebased from #664. Thanks @ejensen!

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/NSImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/NSImage.swift
@@ -118,8 +118,14 @@
       let newRep = NSBitmapImageRep(cgImage: newerCgImage).bitmapData!
       let byteCountThreshold = Int((1 - precision) * Float(byteCount))
       var differentByteCount = 0
-      for offset in 0..<byteCount {
-        if oldRep[offset] != newRep[offset] {
+      // NB: We are purposely using a verbose 'while' look instead of a 'for in' loop.  When the
+      //     compiler doesn't have optimizations enabled, like in test targets, a `while` loop is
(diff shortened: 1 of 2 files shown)


Sources/SnapshotTesting/Snapshotting/NSImage.swift
```diff
@@ -118,8 +118,14 @@
       let newRep = NSBitmapImageRep(cgImage: newerCgImage).bitmapData!
       let byteCountThreshold = Int((1 - precision) * Float(byteCount))
       var differentByteCount = 0
-      for offset in 0..<byteCount {
-        if oldRep[offset] != newRep[offset] {
+      // NB: We are purposely using a verbose 'while' look instead of a 'for in' loop.  When the
```
stephencelis, line 121:
```suggestion
      // NB: We are purposely using a verbose 'while' loop instead of a 'for in' loop.  When the
```

Sources/SnapshotTesting/Snapshotting/UIImage.swift
```diff
@@ -140,8 +140,14 @@
     } else {
       let byteCountThreshold = Int((1 - precision) * Float(byteCount))
       var differentByteCount = 0
-      for offset in 0..<byteCount {
-        if oldBytes[offset] != newerBytes[offset] {
+      // NB: We are purposely using a verbose 'while' look instead of a 'for in' loop.  When the
```
stephencelis, line 143:
```suggestion
      // NB: We are purposely using a verbose 'while' loop instead of a 'for in' loop.  When the
```

stephencelis (approved):

Kaspik (approved):

PR #818, opened 2023-12-14
Remove deprecated UIGraphicsBeginImageContextWithOptions

murtaza64:
As of iOS 17, `UIGraphicsBeginImageContextWithOptions` [has been deprecated](https://developer.apple.com/documentation/uikit/1623912-uigraphicsbeginimagecontextwitho), and causes apps built with Xcode 15 to crash. The suggested fix is to use `UIGraphicsImageRenderer`, which this PR implements.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -182,12 +182,13 @@
     let width = max(old.size.width, new.size.width)
     let height = max(old.size.height, new.size.height)
     let scale = max(old.scale, new.scale)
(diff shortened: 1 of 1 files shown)


adozenlines (changes requested):
These should also consider previous versions of where this API might still be relevant.

murtaza64:
What's the earliest iOS version you need to support? The replacement has been available since iOS 10. Unfortunately I'm not experienced enough with iOS to implement a version check

murtaza64:
Also, there are definitely at least a few other people experiencing crashes:
https://github.com/DylanVann/react-native-fast-image/issues/1010

Issue #780, opened 2023-09-26
Enhancements to `diffTool`?


JackYoustra:
My current workflow for image snapshot tests is usually imagemagik's compare tool, so I'll have a command like
```
compare "1.png" "2.png" png:- | open -a Preview.app -f
```

Unfortunately, the framework only lets me specify the "compare" step instead of providing the rest of the command. I was thinking on changing / adding a closure version of `diffTool` that'd be able to do this. Is there any interest in this or any ideas on how people would like a more generalized `diffTool`?

tisohjung:
I would appreciate it if anyone can review this PR

https://github.com/pointfreeco/swift-snapshot-testing/pull/668

endanke:
@JackYoustra you can probably make an alias to your command which would take the two parameters and run the whole thing.

PR #840, opened 2024-03-23
Add `diffToolBuilder` for building complex diff command messages

js:
This PR adds `SnapshotTesting.diffToolBuilder` closure for building complex diff tool command messages or clickable urls in Xcode. It takes the existing and the failure snapshot as arguments, and thus complex commands which are not just prefixed with the existing `diffTool` can be created. 

For instance, to open [Differati](https://github.com/js/Differati) from a clickable link in Xcode the following `diffToolBuilder` closure can be provided, which URL encodes the paths (in case they contain spaces etc) and builds an url openable by the app:
```swift
SnapshotTesting.diffToolBuilder = { existing, failed in
    func encode(_ path: String) -> String {
        path.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? path
    }
    return "differati://show?old=\(encode(existing.path))&new=\(encode(failed.path))"
}
```

Which gives us an Xcode test failure bubble like this, with a clickable url that takes us directly into the diffing app:
image

Or it can be used to build a complex diff command with the path arguments aren't simply a suffix, as requested in #780:
```swift
SnapshotTesting.diffToolBuilder = { "compare \"\($0.path)\" \"\($1.path)\" png: | open -f -a Preview.app" }
```

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -1,20 +1,28 @@
 import XCTest
 
 /// Enhances failure messages with a command line diff tool expression that can be copied and pasted
-/// into a terminal.
+/// into a terminal. For more complex difftool needs, see `diffToolCommand`.
 ///
 /// ```swift
 /// diffTool = "ksdiff"
 /// ```
-public var diffTool: String? = nil
+public var diffTool: String? {
+    get { diffToolCommand?("", "").trimmingCharacters(in: .whitespaces) }
+    set {
+        diffToolCommand = newValue.map { value in
+            { [value, $0, $1].joined(separator: " ") }
+        }
+    }
+}
 
 /// Enhances failure messages with a diff tool expression created by the closure, such as an clickable
-/// URL or a complex command.
+/// URL or a complex command. The closure will receive the existing screenshot path and the failed
+/// screenshot path as arguments.
 ///
 /// ```swift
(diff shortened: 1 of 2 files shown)


andrewjmeier:
Thank you for doing this!!!

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -323,9 +331,11 @@ public func verifySnapshot<Value, Format>(
       #endif
     }
 
+    let diffToolMessage = diffToolBuilder?(snapshotFileUrl, failedSnapshotFileUrl) ??
+      (diffTool
+        .map { "\($0) \"\(snapshotFileUrl.path)\" \"\(failedSnapshotFileUrl.path)\"" })
     let diffMessage =
-      diffTool
```
stephencelis, line 327: As it stands this PR will break existing use of `diffTool`. We're down to support this feature (as an alternative to #508), but can you make the following change?

1. Update `var diffTool` to be a computed getter/setter to `diffToolBuilder`
2. Consider some alternative names? Maybe `diffToolCommand`?

js: The existing `diffTool` always had precedence over the closure, but agreed a custom getter+setter makes the diffMessage site cleaner. However it wasn't obvious to me how to construct an empty URL for the `diffTool` getter so I changed the `diffToolCommand` closure to receive the path as String rather than the file URL.

Renamed it to `diffToolCommand` instead.

stephencelis (approved):
Thanks! We have an idea to squash things down to a single interface, but will take care of it in another PR.

PR #848, opened 2024-04-16
Add link to plugin SnapshotVision.

gregersson:
A tiny plugin that wraps Apples Vision framework. It allows for snapshotting views based on what texts can be seen. It currently only supports iOS.


mbrandonw (approved):
Very cool, thanks for sharing!

PR #850, opened 2024-04-28
Make ViewImageConfig Sendable

Mika5652:
I got the following warning when I added strict concurrency checking to my project.
Snímek obrazovky 2024-04-28 v 14 10 57
So I decided to suppress them by conforming `ViewImageConfig` to `Sendable` protocol. I hope it's ok to do that.


mbrandonw (approved):
Good catch, thanks!

PR #860, opened 2024-06-18
Add Messaging to Diffing Force Unwrap

nevillco:
Consuming projects can hit the affected force unwrap if there is something wrong with the comparison image file, or if the image file is being stored in `git lfs`. I thought I would offer to include some messaging here. Open to any feedback.


ZevEisenberg:
Hey Connor! I was just running into this due to a misconfigured Git LFS installation, and was thinking it would be nice if this function would fall back to decoding the data as a String and logging it, because it would make the failure so much more obvious at a glance. It could even do an `XCTFail` and/or `Issue.report(...)`, and then return a `UIImage()` so it doesn't crash.

PR #865, opened 2024-07-03
Sketch for supporting Swift Testing

DavidBrunow:
An initial look at what support for Swift Testing could look like. Without this change, tests always pass because `XCTFail` does not fail Swift Testing tests.

Known issues:
* Snapshots are not saved as `XCTAttachments` when tests fail

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -1,5 +1,9 @@
 import XCTest
 
+#if canImport(Testing)
+  import Testing
+#endif
+
 /// Enhances failure messages with a command line diff tool expression that can be copied and pasted
 /// into a terminal.
 ///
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @DavidBrunow, thanks for starting this! 

However, there is still a lot more work to be done. Currently we use test observers in a few spots (e.g. [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/d2341adf332dc7341b4acad8761dff65e905d63c/Sources/SnapshotTesting/AssertSnapshot.swift#L388) and [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/d2341adf332dc7341b4acad8761dff65e905d63c/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift#L278)), and currently swift-testing has no concept whatsoever of test observers. There are plans for it, but in the meantime we are going to have to approximate that concept with "custom execution" traits, which are currently an `@_spi` experimental feature of swift-testing.

We are also _deep_ in the process of updating all of our libraries to be swift-testing compliant, and we have some really exciting things to share soon. But for the time being I think there's too many additional things to think about before we can accept a PR like this. How do you feel about closing this PR and starting a discussion instead?

DavidBrunow:
> Hi @DavidBrunow, thanks for starting this!
> 
> However, there is still a lot more work to be done. Currently we use test observers in a few spots (e.g. [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/d2341adf332dc7341b4acad8761dff65e905d63c/Sources/SnapshotTesting/AssertSnapshot.swift#L388) and [here](https://github.com/pointfreeco/swift-snapshot-testing/blob/d2341adf332dc7341b4acad8761dff65e905d63c/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift#L278)), and currently swift-testing has no concept whatsoever of test observers. There are plans for it, but in the meantime we are going to have to approximate that concept with "custom execution" traits, which are currently an `@_spi` experimental feature of swift-testing.
> 
> We are also _deep_ in the process of updating all of our libraries to be swift-testing compliant, and we have some really exciting things to share soon. But for the time being I think there's too many additional things to think about before we can accept a PR like this. How do you feel about closing this PR and starting a discussion instead?

Done! Sorry for the noise here – I regularly forget about the Discussions feature.

PR #867, opened 2024-07-04
Beta support for Swift Testing, and other improvements.

mbrandonw:
Opening this as a draft so that we can polish it up a bit more.

This PR brings beta support for Swift Testing, and greatly improves the two main customization points for snapshot testing: `diffTool` and `isRecording`. One can now fully customize how the `diffTool` produces a command for two file URLs, and `isRecording` has been turned into an enum with a 3rd option for _never_ recording snapshots (useful for CI) and a forth option for recording only snapshots that fail (useful for strategies that use precision).

Resolves #840, #668, #508, #768, #862, #780, #531, #519


Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -185,174 +199,174 @@ public func verifySnapshot<Value, Format>(
...
-  record recording: Bool = false,
+  record recording: Bool? = nil,
   snapshotDirectory: String? = nil,
   timeout: TimeInterval = 5,
   file: StaticString = #file,
   testName: String = #function,
   line: UInt = #line
 ) -> String? {
-
-  CleanCounterBetweenTestCases.registerIfNeeded()
```
stephencelis, line 196: Do we need to bring this (or something like it) back?

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -185,174 +199,174 @@ public func verifySnapshot<Value, Format>(
...
+        #endif
+      }
 
-    let failureMessage: String
-    if let name = name {
-      failureMessage = "Snapshot \"\(name)\" does not match reference."
-    } else {
-      failureMessage = "Snapshot does not match reference."
-    }
+      // TODO: maybe pass URLs to diffTool so that we can do absolute path for default
```
stephencelis, line 347:
```suggestion
```

stephencelis (approved):

Sources/SnapshotTesting/SnapshotsTestTrait.swift
```diff
@@ -0,0 +1,50 @@
...
+
+  @_spi(Experimental)
+  extension Trait where Self == _SnapshotsTestTrait {
+    /// Configure snapshot testing in a suite or test.
+    /// - Parameters:
+    ///   - diffTool: The diff tool to use in failure messages.
+    ///   - record: The record mode of the test.
+    public static func snapshots(
+      diffTool: SnapshotTestingConfiguration.DiffTool = .default,
+      record: SnapshotTestingConfiguration.Record = .missing
```
stephencelis: @mbrandonw Do you know if traits cascade and if this means these should be optional so that `.missing` won't get applied where a parent set to `.all`?

Sources/SnapshotTesting/SnapshotsTestTrait.swift
```diff
@@ -0,0 +1,52 @@
...
+  @_spi(Experimental)
+  extension Trait where Self == _SnapshotsTestTrait {
+    /// Configure snapshot testing in a suite or test.
+    ///
+    /// - Parameters:
+    ///   - record: The record mode of the test.
+    ///   - diffTool: The diff tool to use in failure messages.
+    public static func snapshots(
+      record: SnapshotTestingConfiguration.Record = .missing,
+      diffTool: SnapshotTestingConfiguration.DiffTool = .default
```
stephencelis, line 13: @mbrandonw Do you know if traits cascade and if this means these should be optional so that `.missing` won't get applied where a parent suite set to `.all`?

PR #869, opened 2024-07-09
Fix recordIssue for Xcode beta 3.

### LINKED ISSUE #868: 1.17.0 doesn't compile on Xcode 16 Beta 3
**Describe the bug**
The library doesn't compile on Xcode 16 Beta 3 due to a change in the swift-testing `Issue.record()` function arguments. Specially the error is: `Extra arguments at positions #2, #3 in call`

The `Issue.record()` function uses `sourceLocation` instead of `filePath` and `line`. It looks like we have to construct an instance of `SourceLocation` and provide in addition to `filePath` and `line` also `column` and `fileID`  the latter can be created using [fileID()](https://developer.apple.com/documentation/swift/fileID())  function so it looks like the `recordIssue()` function of swift-testing will need some changes to pass in an existing SourceLocation or fileID

**To Reproduce**
Open project in Xcode 16 beta 3 and compile.

**Expected behavior**
Compiles ;-)

**Screenshots**

Screenshot 2024-07-09 at 13 24 24

**Environment**
  - swift-snapshot-testing version 1.17.0
  - Xcode 16.0 beta 3 (16A5202i)
  - Swift 6.0
  - OS: 15.0 Beta (24A5279h)

mbrandonw:
Fixes #868 

Xcode beta 3 changed the signature of `Issue.record` to take a source location instead of individual file/line/column.

### DIFF
--- a/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
+++ b/Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
@@ -39,7 +39,7 @@ import Foundation
     timeout: TimeInterval = 5,
     syntaxDescriptor: InlineSnapshotSyntaxDescriptor = InlineSnapshotSyntaxDescriptor(),
     matches expected: (() -> String)? = nil,
-    file: StaticString = #filePath,
+    file: StaticString = #file,
     function: StaticString = #function,
     line: UInt = #line,
     column: UInt = #column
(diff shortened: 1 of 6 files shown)


Sources/InlineSnapshotTesting/AssertInlineSnapshot.swift
```diff
@@ -70,15 +70,29 @@ import Foundation
               loaded. If a timeout is unavoidable, consider setting the "timeout" parameter of
               "assertInlineSnapshot" to a higher value.
               """,
-              file: file,
-              line: line
+              fileID: file,
```
cameroncooke, line 73: According to the header docs `fileID: String` should be a fileID see https://developer.apple.com/documentation/swift/fileid() which is made up of the module name and filename. It can be generated by calling `fileID()` global function from within a macro.

mbrandonw: That would technically be a breaking change and so it will have to be considered more. The test failures created by passing `#file` seem to be ok and so may serve as a stopgap until we are willing to make the breaking change.

cameroncooke: Are there any side effects in terms of where swift testing will show the error location without a valid fileID string, will it show the error at the right point in the source files?

stephencelis (approved):

PR #894, opened 2024-08-24
Fix issue with Testing package in Xcode version 16.0 beta (16A5171c) 

bradhowes:
Using Swift 6 (swiftlang-6.0.0.3.300 clang-1600.0.20.10)

### DIFF
--- a/Sources/SnapshotTesting/Internal/RecordIssue.swift
+++ b/Sources/SnapshotTesting/Internal/RecordIssue.swift
@@ -24,12 +24,10 @@ public func recordIssue(
     if Test.current != nil {
       Issue.record(
         Comment(rawValue: message()),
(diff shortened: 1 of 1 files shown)


mbrandonw:
Hi @bradhowes, the 16A5171c build of Xcode was beta 1, and Swift's Testing framework has gone through a lot of changes since then. We currently only support the newest version of the framework, and it's not really feasible to support older versions. There's no way for us to conditionally provide code depending on the _beta_ of Xcode being used.

If you update to a newer beta things should compile just fine.

Since this isn't an issue with the library I am going to convert this to a discussion. Please feel free to continue the conversation over there.

bradhowes:
Sorry. My fault on that. Thanks!

PR #896, opened 2024-08-27
Run runActivity on main thread (#895)

sk409:
Fixes #895 .


mbrandonw:
Hi @sk409, I believe the only way `runActivity` could be called on a background thread is if your test is running on a background thread. Instead of this change can you mark your test with `@MainActor`?

We have plans to more deeply integrate snapshot testing with Swift's concurrency tools, but until then you will need to be responsible for calling its APIs on the main thread.

sk409:
@mbrandonw
My app uses [SDWebImageSwiftUI](https://github.com/SDWebImage/SDWebImageSwiftUI) to display images, but SDWebImageSwiftUI loads images and updates views on the main thread, so when I run the test on the main thread, XCTWaiter.wait blocks image loading and updating views. Therefore, the image will no longer be displayed in the snapshot.
For the reasons mentioned above I need to run my tests on a background thread instead of the main thread.

PR #916, opened 2024-10-06
Fix assertSnapshot for Swift Testing tests.

### LINKED ISSUE #915: Crash `Current context must not be nil` with swift-testing
**Describe the bug**
I am running into an uncaught exception “Terminating app due to uncaught exception `NSInternalInconsistencyException`, reason: `Current context must not be nil`” when snapshot testing a string. _The problem only occurs when the test fails_. There is no problem at all when recording.

This is not an async test.

**To Reproduce**

Fresh new project

```swift
import Testing
import SnapshotTesting

struct SomeTests {
    @Test
    func this_test_crashes_when_assertion_fails() throws {
        withSnapshotTesting {
            assertSnapshot(of: "foo", as: .lines)
        }
    }
}
```

**Expected behavior**
Maybe I am missing something fundamental in the new snapshottesting setup, but this should not crash and I should get a diff in Xcode.

**Screenshots**
Similar screenshot as in #875 

**Environment**
  - swift-snapshot-testing version 1.17.5
  - Xcode 16.0 (16A242d)
  - Swift 6.0
  - OS: macOS 15.0.1 (24A348)

**Additional context**
This is not an async test, so #822 and #875 solutions do not apply here

Thank you in advance for looking into this!

mbrandonw:
Fixes #915

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,10 +13,11 @@ jobs:
     strategy:
       matrix:
         xcode:
-          - "14.3.1"
+          - 15.4
+          - '16.0'
 
-    name: macOS 13 (Xcode ${{ matrix.xcode }})
-    runs-on: macos-13
+    name: macOS
+    runs-on: macos-14
     steps:
       - uses: actions/checkout@v3
       - name: Select Xcode ${{ matrix.xcode }}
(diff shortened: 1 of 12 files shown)


stephencelis (approved):

IlyaPuchkaTW:
Does this mean that attachments won't be available at all when running from Swift Testing?

stephencelis:
@IlyaPuchkaTW That's correct. You can follow their progress here: https://github.com/swiftlang/swift-testing/issues/714

PR #925, opened 2024-11-03
Compile for Android

johankool:


johankool:
@marcprux did it better. 😄

PR #939, opened 2025-01-05
Android support

marcprux:
This PR adds support for Android and a CI action that tests against the Android emulator using [swift-android-action](https://github.com/marketplace/actions/swift-android-action).

It addresses the issue with https://github.com/pointfreeco/swift-snapshot-testing/pull/925 by using the `copy-files` parameter to copy the `__Snapshots__` folder up to the emulator, and then special-cases the Android tests to look in that folder when running the snapshot tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,15 @@ jobs:
       - run: swift test
 
   android:
+    strategy:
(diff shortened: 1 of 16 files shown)


stephencelis (approved):
Thanks!

PR #943, opened 2025-01-14
Record Attachments in Never mode

jflan-dd:
Add XCTAttachment even for `.never` record mode

Context:
https://github.com/pointfreeco/swift-snapshot-testing/discussions/940


mbrandonw:
Hey @jflan-dd, thanks for looking into this!

To get CI passing I think you will need to bracket off some of the new code in a `#if !os(Linux) && !os(Windows)` check. Also, GitHub has sunset Xcode 16.0, so could you update ci.yml to use 16.1?

jflan-dd:
Thanks @mbrandonw!

CI is passing now

mbrandonw:
@jflan-dd, great, thanks again!

Issue #805, opened 2023-11-08
Timeout Error During Snapshot Testing with WKWebView


stephan-vivie-rb:
## Environment
- Xcode Version: 15.0.1
- Swift Version: 5.9
- SnapshotTesting Version: 1.14.0
- iOS Version: 17.0.1
- Devices/Emulators: iPhone 15 iOS 17.0.1

## Description
We are experiencing a timeout error when performing snapshot tests on a `UIViewController` which contains a `WKWebView`. Our tests consistently fail with the following error message:

```
failed - Exceeded timeout of 5.0 seconds waiting for snapshot.

This can happen when an asynchronously rendered view (like a web view) has not loaded. Ensure that every subview of the view hierarchy has loaded to avoid timeouts, or, if a timeout is unavoidable, consider setting the "timeout" parameter of "assertSnapshot" to a higher value.
```

I have tried to increase the timeout but that doesn't help. 

Is there a recommended approach for ensuring the WKWebView has fully rendered before the snapshot is taken? If that's not possible. It would be great just to be able to run the snapshot tests on the rest of the screen as the content of the WKWebView is out of our hands anyway, but the other elements on screen are not.

haemi:
did you find a solution for this? I'm experiencing the same

pewkachu:
yay, at least I'm not alone with this 🎉

stephan-vivie-rb:
> did you find a solution for this? I'm experiencing the same

Unfortunately not.

johnpatrickmorgan:
Please check this fork and see if it fixes the issue for you: https://github.com/johnpatrickmorgan/swift-snapshot-testing/tree/fix-ios-wkwebview-timeout.

PR #944, opened 2025-01-17
WKWebView timeout: ensure subscription is not invalidated until isLoading becomes false

johnpatrickmorgan:
This PR is a fix for #805. 

The current logic stops observing `wkWebView`'s `isLoading` property as soon as the observation handler is called. If the observation is first called when `isLoading` is `true`, the observation is invalidated and the handler will not be called again when `isLoading` flips to false, which leads to timeout errors on iOS. For some reason, on macOS, invalidating the subscription does not seem to prevent the handler being called again once `isLoading` changes to false.

Note: I'd like to be able to verify this fix with tests, but I get transparent images returned on iOS when snapshotting a `WKWebView`. This seems to be related to #857, but even the suggested fix there does not overcome the issue for me.

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -889,10 +889,10 @@
               var subscription: NSKeyValueObservation?
               subscription = wkWebView.observe(\.isLoading, options: [.initial, .new]) {
                 (webview, change) in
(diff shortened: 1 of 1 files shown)


OguzYuuksel:
Hi @stephencelis  & @mbrandonw, is it possible to review & merge that?

Issue #919, opened 2024-10-17
swift 6.0.1 crash when running tests swift testing with @MainActor annotation only on linux


doozMen:
**Describe the bug**
Only happens on linux 
1. clone https://github.com/doozMen/swift-testing-main-bug
2. run unit tests with visual studio in debug

You will see it crashes. The docs state that `Thread.isMainThread` is not available. Weirdly it is on macOS just not on linux. 

**To Reproduce**
run the test project on linux and maybe add code 

```swift
    #expect(Thread.isMainThread == true)
```

This will work on mac and not on linux

**Expected behavior**

Not creating deadlock because of doing dispach on main in sync from main

**Screenshots**
<img width="1303" alt="Screenshot 2024-10-17 at 16 16 31" src="https://github.com/user-attachments/assets/56f153d3-e1b3-4b3c-b109-7e739adab3c3">

**Environment**
  - swift-snapshot-testing version [1.17..5]
  - Swift [6.0.1]
  - OS: aarch64-unknown-linux-gnu

**Additional context**
Maybe also a bug in swift but not sure as when you the check is unavailable in async context according to the docs. So you should not rely on it?

Maybe have a asser option that is marked with @MainActor might be better?

stephencelis:
@doozMen Thanks for the report. We're running SnapshotTesting on Linux for pointfree.co, but are on a particular branch that uses this logic to dispatch to the main queue instead:

https://github.com/pointfreeco/swift-snapshot-testing/blob/7de6fa587c7fa38e435186e124f1d8c11d36ceb3/Sources/SnapshotTesting/AssertSnapshot.swift#L352-L359

Would you be open to exploring if this fix would address the problem for you when applied to the `main` branch?

peterkovacs:
This issue is also affecting me in a very similar way. [My project](https://github.com/peterkovacs/swift-tui) is using swift 6, and swift-testing exclusively. All my suites are marked as @MainActor and I get the following crash at run-time:

```
*** Signal 5: Backtracing from 0xffffa84b3e98... done ***

*** Program crashed: System trap at 0x0000ffffa84b3e98 ***

Thread 0 crashed:

 0                  0x0000ffffa84b3e98 __DISPATCH_WAIT_FOR_QUEUE__ + 352 in libdispatch.so
 1 [ra]             0x0000ffffa84b38d4 _dispatch_sync_f_slow + 163 in libdispatch.so
[...]

PR #946, opened 2025-01-23
Use alternate means to detect if running on main thread

peterkovacs:
This is to avoid problems on linux when running `@MainActor` tests.

Resolves issue #919


stephencelis (approved):
Thanks!

PR #948, opened 2025-01-24
Add `.customDump` strategy and deprecate `.dump`

stephencelis:
Custom Dump already does a lot of the work the `.dump` strategy does, but also a whole lot more and in a friendlier format.


stephencelis:
- [ ] Add test

PR #952, opened 2025-02-04
Disable XCTAttachment when using Swift Testing

ralfebert:
Adding the XCTAttachments for UIImage comparisons seems to be way more expensive than the image comparison itself:

Screenshot 2025-02-04 at 10 41 59

I suggest not adding them when using Swift Testing.


jflan-dd:
We heavily rely on having recorded snapshots bundled in XCResult files. Please at least make this an opt-in/opt-out change.

stephencelis:
We don't plan on changing this at the moment, and Swift Testing attachments are in active review: https://forums.swift.org/t/st-0009-attachments/78698

PR #962, opened 2025-03-26
Fix potential deadlock

### LINKED ISSUE #961: Deadlock when `assertSnapshot` called on main thread but not on main queue
**Describe the bug**

[This commit](https://github.com/pointfreeco/swift-snapshot-testing/pull/946) changes the approach to detect if a snapshot is generated on the main thread. However, as discussed in [GRDBSnapshotTests issue](https://github.com/groue/GRDBSnapshotTesting/issues/5), this is a breaking change for certain code patterns.

**To Reproduce**

Download [GRDBSnapshotTesting](https://github.com/groue/GRDBSnapshotTesting) and add the following test case:

```
// Crash
func test_issue_5() throws {
    let dbQueue = try DatabaseQueue()
    try dbQueue.read { db in
        assertSnapshot(of: "SELECT 'Hello, world!'", as: .dump(db))
    }
}
```

Or: 
```swift

// On the main thread
let queue = DispatchQueue(label: "test")
queue.sync {
    print("Hello")

    // Success: we're on the main thread
    assert(Thread.isMainThread)

    // assertSnapshot ...
}
```

**Expected behavior**
A snapshot is successfully recorded and generated

**Environment**
  - swift-snapshot-testing version 1.18.0 and higher
  - Xcode [e.g. 16.2]
  - Swift [e.g. 6]
  - OS: [e.g. iOS 18]

stephencelis:
Unfortunately #946 introduced a deadlock when `assertSnapshot` is called on the main thread but not the main queue, so let's revert to the old code path for now with an explicit `Test.current` check to avoid the original issue.

Fixes #961.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -285,7 +285,13 @@ public func verifySnapshot<Value, Format>(
   line: UInt = #line,
   column: UInt = #column
 ) -> String? {
-  CleanCounterBetweenTestCases.registerIfNeeded()
+  #if canImport(Testing)
+    if Test.current == nil {
+      CleanCounterBetweenTestCases.registerIfNeeded()
+    }
+  #else
+    CleanCounterBetweenTestCases.registerIfNeeded()
+  #endif
 
   let record =
     (recording == true ? .all : recording == false ? .missing : nil)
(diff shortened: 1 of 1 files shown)


stephencelis:
@jcavar Can you please run your tests against this branch?

@peterkovacs @doozMen Can you please verify that the original problem is still addressed?

peterkovacs:
I can't recall if it was a crash that would happen with every test run, but I've run with this branch a few dozen times and all runs were successful.

PR #990, opened 2025-05-21
Help with file paths that include spaces

FishOSU:
The error messages for failing tests include ksdiff output to copy/paste into the command line. If you happen to have spaces anywhere in your path structure it fails. This just wraps them so that it launches Kaleidoscope as expected.


mbrandonw (approved):
Hi @FishOSU, good catch, thanks!

PR #992, opened 2025-05-28
Using explicit .git URLs for dependencies and installation steps

okalentiev:
Thank you for a great library. 

We are experiencing a tiny annoyance with package resolution. Occasionally, when resolving packages, Xcode would use different URLs for resolution, resulting in constant back-and-forth changes to the `Package.resolved` file. 

```diff
{
       "identity" : "swift-syntax",
       "kind" : "remoteSourceControl",
-     "location" : "https://github.com/swiftlang/swift-syntax.git",
+     "location" : "https://github.com/swiftlang/swift-syntax",
```

According to my research, this is an Xcode annoyance where, depending on the local setup or mood 🤷, it would select one URL or another. The solution involves unifying the URLs defined by the project and its dependencies to use a single format. I have checked the installation instructions for swift-syntax, and they suggest using URLs with .git at the end, hence this change. 

I considered starting a conversation about this, but the change is minor, so I decided to do a PR instead. I welcome the conversation and understand that everyone's setup and environment are different, and such a change can be pretty invasive. 

Resources: 
https://forums.swift.org/t/xcode-automatically-updating-package-resolved/41900/6
https://forums.swift.org/t/is-there-is-a-good-reason-to-include-git-in-swift-package-dependency-urls/68800
https://www.reddit.com/r/swift/comments/193opr4/spm_randomly_adds_and_deletes_git_extension_from/

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,4 +1,5 @@
 {
+  "originHash" : "07a521480635f19eb7317262fb971d3ab6c7aab59602add7dd5b3f44a72b1ef7",
   "pins" : [
     {
       "identity" : "swift-custom-dump",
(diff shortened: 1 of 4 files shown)


okalentiev:
@stephencelis Thanks for your response! 

While SwiftPM itself normalises URLs with and without `.git`, Xcode sometimes rewrites them in `Package.resolved`. I agree that enforcing a single URL style project-wide would only be beneficial if everyone adhered to the same standard, but as you pointed out, many projects use both forms. Indeed, chasing consensus across the ecosystem isn’t practical, so the real fix needs to come from Apple. 

I haven’t filed feedback yet, as I'm struggling with creating a reproducible project. I'm not sure if this is caused by the combination of local cache + switching branches or some other Xcode quirk. If I get to it, I will ping you. I'll close the PR for now.

Thanks for looking into this!

PR #998, opened 2025-06-07
Relax the protocol requirement for Snapshotting extension

Kyle-Ye:
When wrapping a custom assert function for SwiftUI.View, we may want to write the following code.

```
#if canImport(AppKit)
typealias PlatformHostingController = NSHostingController
#else
typealias PlatformHostingController = UIHostingController
#endif

func viewAssertSnapshot<V: View, Format>(
    of value: @autoclosure () throws -> V,
    as snapshotting: Snapshotting<PlatformHostingController<V>, Format> = .image,
    ...
) {
    assertSnapshot(
        of: PlatformHostingController(rootView: try value()),
        as: snapshotting,
        named: name,
        record: recording,
        timeout: timeout,
        fileID: fileID,
        file: filePath,
        testName: testName,
        line: line,
        column: column
    )
    ...
}
```

It does not compile now since image extension has a requirement of `Value == UIViewController`.

And the workaround is change `Snapshotting<PlatformHostingController<V>, Format>` to `Snapshotting<UIViewController, Format>`.

I think it is common we define some UIViewController's subclass and use the new subclass as the generic. So I propose we relax the requirement for the common UI extension of Snapshotting.

### DIFF
--- a/Sources/SnapshotTesting/Documentation.docc/Articles/CustomStrategies.md
+++ b/Sources/SnapshotTesting/Documentation.docc/Articles/CustomStrategies.md
@@ -22,7 +22,7 @@ Snapshotting<UIView, UIImage>.image
 We can define an `image` strategy on `UIViewController` using the `pullback` method:
 
 ``` swift
-extension Snapshotting where Value == UIViewController, Format == UIImage {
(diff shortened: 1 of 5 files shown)


Kyle-Ye:
I will rebase it locally today to reproduce and fix the CI issue

PR #1006, opened 2025-06-28
Fixes #1005, trailing trivia preserved when replacing closures

### LINKED ISSUE #1005: Multiple Inline Snapshots Do Not Preserve Trailing Trivia When Replacing Closures
When using multiple inline snapshots, trailing trivia is not preserved when replacing closures.

**To Reproduce**
Change text in the `head:` closure of `assertResponse` in multipleInlineSnapshots() AssertInlineSnapshotSwiftTests.swift:154 and run the test. Notice the space after the closure, before the `body:` label is removed.

**Expected behavior**
Expect trailing trivia to preserved (space should not be removed).

seanwoodward:
Fixes #1005


stephencelis (approved):
Great catch, thanks!

PR #1028, opened 2025-09-25
Improvements to the generated difference image (#520)

### LINKED ISSUE #520: Difference image doesn't show difference in opacity.
If the difference in two images is only the alpha channel but the rgb channels are the same, the difference image won't show that anything has changed. 

## Example
In this case it's a `UILabel` that has where the textColor is black (`rgba(0, 0, 0, 1)`) and the backgroundColor is clear (`rgba(0, 0, 0, 0)`).

Reference

Failure

Difference

If I take the reference and failure images and pass them to ImageMagick's `compare` command, it gives the following image showing the differences..

If I change view being snapshot to have a white background, then the created difference image shows the changes. While this does work, adding a background color to the view then means it isn't exactly the same as the view that'll be used in the app.

CraigSiemens:
This PR improves the generated difference image to make differences much clearer in cases where he alpha component changes or there's only a subtle change in color.

- Fixes #520
- Now shows differences when only the alpha component changes
- Normalizes the difference values to improve contrast in the image
- Falls back to using the difference blend mode when the image size has changed

# Examples

All the images below contain the "reference" "failure" and "difference" images from one test run.

## Alpha only changes

The "reference" is a transparent image containing the word "Hello"
The "failure" is a transparent image with the same size containing the word "World"

### Before

The "difference" image is all black because the alpha component is not considered by the `difference` blend mode.

before

### After

The "difference" image now highlights pixels with changes to their the alpha component.

after

## Subtle color changes

The "reference" image is the color `Color(white: 128.0 / 255.0)`
The "failure" image adds the text "Changed" with the foreground style `Color(white: 129.0 / 255.0)`

### Before

The "difference" is appears all black. The text is so close to the background color, it
s in the "difference" image with a color of `Color(white: 1.0 / 255.0)` which easily not perceptible. You need to adjust the levels in an image editor to really make them visible.

before

### After

The "difference" image now clearly shows the difference by normalizing the difference values in the image based on the max component difference per pixel. White is the most different pixel in the image, and black is unchanged pixels.

after

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -179,6 +179,11 @@
   }
 
   private func diff(_ old: UIImage, _ new: UIImage) -> UIImage {
+    normalizedComponentDiff(old, new)
+    ?? blendModeDiff(old, new)
+  }
+
+  private func blendModeDiff(_ old: UIImage, _ new: UIImage) -> UIImage {
     let width = max(old.size.width, new.size.width)
     let height = max(old.size.height, new.size.height)
     let scale = max(old.scale, new.scale)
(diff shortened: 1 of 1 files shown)


Kaspik:
@stephencelis Please please, get this merged, it's been forever and current version is just wrong.

stephencelis (approved):
@CraigSiemens Thanks for this! We think it should be an improvement. I was a little worried about performance changes but it's probably not a big deal for the unhappy path.

PR #1035, opened 2025-11-10
Update package versioning

stephencelis:
SPM recommends `Package.swift` point to the latest version, but we're doing the opposite.

### DIFF
--- a/Tests/InlineSnapshotTestingTests/AssertInlineSnapshotSwiftTests.swift
+++ b/Tests/InlineSnapshotTestingTests/AssertInlineSnapshotSwiftTests.swift
@@ -28,15 +28,15 @@
             """
           }
         } matching: { issue in
-          issue.description == """
-            Issue recorded: Snapshot did not match. Difference: …
+          issue.description.hasSuffix("""
+            Snapshot did not match. Difference: …
 
               @@ −1,3 +1,4 @@
                ▿ 2 elements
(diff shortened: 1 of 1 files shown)


mbrandonw (approved):

stephencelis:
@marcprux Do you know if there are any changes to the Android action you contributed that may have led to it start failing here? Both this PR and a PR on its dependency, https://github.com/pointfreeco/swift-macro-testing/pull/51 have started to see failing Android CI.

stephencelis:
@marcprux Going to disable Android CI for now but will happily take a PR to restore it in the future!

marcprux:
Hrm … nothing should have changed there. That linking error looks vaguely familiar, but I don't recall exactly when I've seen it.

I'll look into it and file a PR if I can get it working on a fork. I suspect that updating the SDK from 6.0.2 to something more recent (like 6.2) might help.

marcprux:
I tried `swift-macro-testing` in a matrix with 3 swift versions (6.0.3, 6.1, and 6.2) at https://github.com/swift-everywhere/swift-macro-testing/actions/runs/19377036667, and the 6.0.3 version is the only one that fails. Hopefully using the more recent version will work for you, but I'll also be looking into the 6.0.3 failure, since there's no reason I can identify why that should suddenly have stopped working.

PR #1049, opened 2026-01-29
Revert "Don't install test observer when in Swift Testing test (#1048)"

mbrandonw:
This reverts commit cc051f1c07a414cf36b3d45fc8d00f2a8086ca5e.

This change aimed to fix inline snapshots on Linux, which we still want to do, but it's going to require a bigger change.


mbrandonw:
Closing in favor of #1050

PR #1051, opened 2026-02-03
Use windows from UISceneDelegate lifecycle

opiotr:
###  Problem
After migrating to UISceneDelegate lifecycle, snapshot tests failed with pixels mismatch due to incorrect safe area insets being applied. Also other sizing issues appeared which caused fails for currently recorded view's snapshots.

###  Root Cause
The internal Window class was not attached to the active UIWindowScene, causing inconsistent behavior when rendering  views.

### Proposed changes
1. Updated `getKeyWindow()` method
   - Use scene-based window lookup (connectedScenes) on iOS 15+ instead of deprecated UIApplication.shared.windows
2. Updated `Window` class initializer
   - Attach snapshot window to the active UIWindowScene on iOS 13+

### DIFF
--- a/Sources/SnapshotTesting/Common/View.swift
+++ b/Sources/SnapshotTesting/Common/View.swift
@@ -1076,7 +1076,12 @@
 
     private func getKeyWindow() -> UIWindow? {
       var window: UIWindow?
-      if #available(iOS 13.0, *) {
+      if #available(iOS 15.0, *) {
+        window = UIApplication.sharedIfAvailable?.connectedScenes
+          .compactMap { $0 as? UIWindowScene }
+          .flatMap { $0.windows }
+          .first { $0.isKeyWindow }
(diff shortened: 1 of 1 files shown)


Sources/SnapshotTesting/Common/View.swift
```diff
@@ -1090,7 +1097,19 @@
       init(config: ViewImageConfig, viewController: UIViewController) {
         let size = config.size ?? viewController.view.bounds.size
         self.config = config
-        super.init(frame: .init(origin: .zero, size: size))
+
+        // Attach to current window scene to ensure consistent safe area behavior
+        if #available(iOS 13.0, *),
+          let scene = UIApplication.sharedIfAvailable?.connectedScenes.first(where: {
```
Ernest0-Production, line 1103: This doesn't seem to be the safe solution, as it forces the host app to be connected to the test target.

PR #1067, opened 2026-03-19
Prefer native Swift Testing image attachments on supported toolchains

ajkolean:
This builds on the Swift Testing attachment support added in #1064 and refined in #1065.

Currently, Swift Testing attachments in `AssertSnapshot.swift` are recorded as generic `Data` attachments via `Attachment.record(data, ...)`. On newer toolchains, Swift Testing supports attaching image types directly, so PNG snapshot artifacts can be recorded as native image attachments instead.

This change adds a small helper to prefer native Swift Testing image attachments for PNG snapshot artifacts when supported, while preserving the existing fallback behavior everywhere else.

### What changes

- add a small `recordSwiftTestingAttachment` helper
- for `.png` attachments on supported toolchains:
  - decode the data as `UIImage` / `NSImage`
  - record it with `Attachment.record(..., as: .png, sourceLocation: ...)`
- otherwise fall back to the existing `Attachment.record(data, ...)` path
- update both Swift Testing attachment recording sites to use the helper
- preserve existing `SourceLocation` behavior from #1065

### Notes

- This is a focused enhancement to the existing Swift Testing attachment path.
- Non-image attachments are unchanged.
- Older toolchains continue to use the existing `Data` attachment behavior.
- XCTest behavior is unchanged.

Xcode 26.4 adds native Swift Testing image attachments, so this change aligns SnapshotTesting’s PNG snapshot artifacts with the newer test-reporting API on supported toolchains.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -9,35 +9,6 @@ import XCTest
 
 #if canImport(Testing)
   import Testing
-
-  #if compiler(>=6.2)
-    private func recordSwiftTestingAttachment(
-      _ data: Data,
-      named name: String,
-      sourceLocation: SourceLocation
-    ) {
-      #if compiler(>=6.3)
-        if name.hasSuffix(".png") {
-          #if canImport(UIKit)
-            if #available(iOS 14.0, tvOS 14.0, *) {
-              if let image = UIImage(data: data) {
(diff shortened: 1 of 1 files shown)


stephencelis (approved):
Thanks!

gohanlon:
Hi @ajkolean,

I'm trying to run down the right fix for #1085. Can you clarify what motivated this PR? Was it to fix the build, given the new `Attachable` protocol requirement? Or does Xcode 26.4 do something with the new native Swift Testing image attachments overloads that it doesn't do with raw data?

PR #1084, opened 2026-04-15
Fix color space issue for snapshot with drawHierarchyInKeyWindow set to true

### LINKED ISSUE #1083: Color space mismatch when comparing snapshots with drawHierarchyInKeyWindow: true
**Describe the bug**
Give a clear and concise description of what the bug is.

In one of the projects I have noticed issues with snapshot tests that uses `.image(drawHierarchyInKeyWindow: true)` snapshotting option.
Further investigation shows that snapshot create with this option inherits P3 color spaces as compared to sRGB used in other cases.
When running this test again on Mac with same GPU architecture all works fine. Issues occurs when such snapshot is evaluated with Mac using different GPU architecture (eg M2 vs M4) or there is slight change to actual snapshotted image.

In case of GPU architecture changed: test recorded on M2 Mac fails when run on M4 mac.
Example failure: 
```
The percentage of pixels that match 0.0859375 is less than required 1.0
The lowest perceptual color precision 0.83484375 is less than required 0.999
```

In case of any tiny change to snapshot as compared to reference it will certainly fail (almost all pixels differs).
Example failure (single pixel changed from black to grey):
```
The percentage of pixels that match 0.0859375 is less than required 1.0
The lowest perceptual color precision 0.77796876 is less than required 0.999
``` 
Whereas similar view captured using `.image(drawHierarchyInKeyWindow: false)` fails with:
```
The percentage of pixels that match 0.99999833 is less than required 1.0
The lowest perceptual color precision 0.77750003 is less than required 0.999
```

See `percentage of pixels that match` in examples above. For both changed GPU architecture or tiny change almost all pixels differ (0.0859375 match).

**To Reproduce**

Project below shows case described above as "tiny change"
[Snap bug color space.zip](https://github.com/user-attachments/files/26756574/Snap.bug.color.space.zip)
Same project can be found in this repo: https://github.com/TomaszLizer/SnapshotTesting-color-space-issue

To reproduce issue run tests and inspect failure description for first test, compare it with second one (both have same change made to reference).

**Expected behavior**
Snapshotting is less sensitive to tiny changes when using .image(drawHierarchyInKeyWindow: true)` snapshotting option.

**Screenshots**

Image

Image

**Environment**
  - swift-snapshot-testing version: 1.19.2
  - Xcode: 26.4
  - Swift: 5 
  - OS: iOS 26.4 (iPhone 17 simulator)

TomaszLizer:
Use  CGImage with normalised color space for perceptual comparison.

Partially resolves  https://github.com/pointfreeco/swift-snapshot-testing/issues/1083
There is still issue when running tests with same env but on different GPU architecture (eg M2 Mac vs M4 Mac), yet this can be handled much easier with high value of perceptual precision as compared to current state.

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -110,7 +110,8 @@
     let pixelCount = oldCgImage.width * oldCgImage.height
     let byteCount = imageContextBytesPerPixel * pixelCount
     var oldBytes = [UInt8](repeating: 0, count: byteCount)
-    guard let oldData = context(for: oldCgImage, data: &oldBytes)?.data else {
+    guard let oldContext = context(for: oldCgImage, data: &oldBytes),
+          let oldData = oldContext.data else {
       return "Reference image's data could not be loaded."
     }
     if let newContext = context(for: newCgImage), let newData = newContext.data {
@@ -130,9 +131,12 @@
       return "Newly-taken snapshot does not match reference."
     }
     if perceptualPrecision < 1, #available(iOS 11.0, tvOS 11.0, *) {
+      guard let oldNormalized = oldContext.makeImage(),
+            let newNormalized = newerContext.makeImage()
+      else { return "Snapshot color space normalization failed." }
       return perceptuallyCompare(
-        CIImage(cgImage: oldCgImage),
-        CIImage(cgImage: newCgImage),
+        CIImage(cgImage: oldNormalized),
+        CIImage(cgImage: newNormalized),
         pixelPrecision: precision,
         perceptualPrecision: perceptualPrecision
       )


knutanderss:
Just curious about this fix, since we're experiencing problems related to color space on our team (otherwise very happy with this library!). What exactly is solved and what is not solved with this PR? 😊

TomaszLizer:
> Just curious about this fix, since we're experiencing problems related to color space on our team (otherwise very happy with this library!). What exactly is solved and what is not solved with this PR? 😊

There are two issues that we are having on our side:
1. Color space mismatch
2. Different GPU architecture - rendering minute differences in snapshots rendered using `.image(drawHierarchyInKeyWindow: true)` option.

This PR fixes first issue.
Existing code normalizes image when performing byte level check - hence this is not an issue if there is no change in the rendered output. If there is however minute difference, then per pixel checks kicks in. Color space mismatch causes tiny changes in actual values to explode by orders of magnitude.
Imagine you have some test, that you set to it `perceptualPrecission` so that you don't need to re0record on eg SDK update. However here - because color space changes - even minute change in render causes almost whole comparison to fail, eg in some of tests we have: `The percentage of pixels that match 0.0859375 is less than required 1.0`.

On top of that issue - which causes perceptualPrecission to be unusable for tests using `drawHierarchyInKeyWindow` we have in our team different machines across the team:
- M1 based CI
- M2 based MacBooks 
- M4 based MacBooks 

Now there was architectural change of GPU between M2 - M3 and it can cause tiny precession errors when computing same data. Because of that, tests that are recorded on eg M2 - never passes on M4 because there are tiny differences, then those differences are magnified by color space issue causing total mismatch and making it very hard to work with.

So this PR fixes color space mismatch.
It does not tackle however lossy creation of the snapshot when using `drawHierarchyInKeyWindow`. I guess there might be some option, to persist those snapshot in a way, that byte comparison will render them the same, but I expect some GPU operations to be made on the image before save that causes those differences.
It does not matter that much however and this fix together with high perceptual precision (eg 0.995) solves issue for us.

TomaszLizer:
@stephencelis @mbrandonw Would you mind having a look :) 
Fairly simple change, but of substantial value in my team. Right now we run a fork with this fix to mitigate this issue.

stephencelis:
@TomaszLizer If only simple changes were simple in this library :)

The problem is that many hundreds of thousands of tests in the wild depend on the existing strategy's implementation, and even the smallest change can break an existing suite. The original strategy passed the original color space, but was normalized to sRGB back in #446, and a few more recent PRs have made small changes from there.

Your change would seem to revert that behavior (though you use `makeImage()` instead, which may result in a more forgiving change), and so it might break existing suites that rely on the original PR.

Our plan in 2.0 is to ship a far simpler image snapshot strategy that doesn't try to accommodate every edge case, and then folks can maintain their own forks of this strategy if they have more specific requirements.

So I guess my main question is with regards to this:

> Right now we run a fork with this fix to mitigate this issue

Is there a reason you must maintain an entire fork rather than use the main library but maintain just the image strategy?

TomaszLizer:
Hey @stephencelis 

> Your change would seem to revert that behavior (though you use makeImage() instead, which may result in a more forgiving change), and so it might break existing suites that rely on the original PR.

Could you help me understanding how my changes reverts those made in mentioned PR?
If I understand that correctly, then my changes not only do not revert those changes, but extends them to pixel comparison. Especially mentioned PR tackles bytes comparison using normalised context, whereas mine handles later check - pixel comparison.

Currently if bytes checks failed - we perform pixel checks -> non-normalised for color spaces. This causes in some circumstances comparison to be made using different color-space for snapshot and reference which obviously cannot succeed.
For reference you can check demo repo attached to the issue, where changing single pixel causes snapshot to fail substantially (almost no pixel matches).

> Is there a reason you must maintain an entire fork rather than use the main library but maintain just the image strategy?

To me it was obvious that sharing fix for such flawed check is beneficial to others, isn't it? Also current strategies suits me well, so didn't saw a reason to create whole new one :)

stephencelis:
> Could you help me understanding how my changes reverts those made in mentioned PR? If I understand that correctly, then my changes not only do not revert those changes, but extends them to pixel comparison. Especially mentioned PR tackles bytes comparison using normalised context, whereas mine handles later check - pixel comparison.

This may just be my distance from the problem and a lack of my understanding of the changes. My initial worry was that `makeImage()` was resurrecting the original color space and would run counter to the original PR. Are you suggesting that on failure the normalized image context must invoke `makeImage()` to further normalize the underlying bytes to the same sRGB content?

We're definitely open to the change, though I imagine `makeImage()` is not free and so failures may lead to substantially more work, and a big UI change could slow down a test suite...

Maybe @dflems could also point his suite to this PR to confirm it doesn't break anything?

TomaszLizer:
> This may just be my distance from the problem and a lack of my understanding of the changes. My initial worry was that `makeImage()` was resurrecting the original color space and would run counter to the original PR. Are you suggesting that on failure the normalized image context must invoke `makeImage()` to further normalize the underlying bytes to the same sRGB content?

@stephencelis Exactly that - current comparison is not using any normalization causing failures for pixel comparison (perceptually compare) for some strategies (those that provides image with different color space).
By adding `makeImage()` I reuse same normalization. Also when we enter this check we already exhausted lightweight checks (raw bytes comparison + normalized bytes comparison) and this check itself is already on the expensive side (comparing each pixel one by one).

Personally I would expect such normalization to have no big effect, but can measure if necessary.

Alternatively can see two other options:
- Similarly to byte check: perform raw check first (on cgImage provided as is) and fallback to normalized check if it fails
- Verify color space beforehand and normalize images if those do not match.

Some performance testing would be nice, especially if someone already have a testing suite for that (@dflems)?
My gut feeling was just that at this stage there is not much to loose or gain anyway so did not bother to go deeper.

TomaszLizer:
@stephencelis What is status of the review? :) 
Do we wait for @dflems to chime in, or maybe I should make some benchmark / changes on my side already?

TomaszLizer:
@stephencelis Ping :)

PR #1088, opened 2026-04-24
Document .snapshots trait requirement for Swift Testing

PSinha1202:
## Summary

- Adds the `.snapshots` trait to the primary Swift Testing usage example in the README
- Adds an **Important** callout explaining why the trait is required and what happens without it (flaky failures, duplicate snapshot recordings)
- Updates the Features list from "No configuration required" to "Minimal configuration" with a pointer to the usage section

## Context

When using Swift Testing (not XCTest), the `.snapshots` trait must be applied to `@Suite` or `@Test` for the library's internal snapshot counter to reset between test iterations. Without it, tests can fail unexpectedly or flake — especially when running with `xcodebuild -test-iterations` or repeated Xcode runs. This was observed in production CI pipelines and discussed in #995.

XCTestCase-based tests are unaffected because `invokeTest()` manages the lifecycle automatically.

## Test plan

- [x] Verified README renders correctly with the new callout
- [x] No code changes — documentation only

Ref: https://github.com/pointfreeco/swift-snapshot-testing/discussions/995#discussioncomment-13313498

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,15 @@ Delightful Swift snapshot testing.
 
 ## Usage
 
-Once [installed](#installation), _no additional configuration is required_. You can import the
-`SnapshotTesting` module and call the `assertSnapshot` function.
+Once [installed](#installation), you can import the `SnapshotTesting` module and call the
+`assertSnapshot` function.
 
 ``` swift
 import SnapshotTesting
 import Testing
 
 @MainActor
+@Suite(.snapshots)
 struct MyViewControllerTests {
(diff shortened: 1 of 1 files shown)


hanneskaeufler:
Ah, this would be a good addition, I scratched my head for quite a while on flaky results until I randomly saw this today.

davdroman:
Yeah this is essential guidance imo, should be more prominent. @stephencelis can you take a look please? :)

PR #1098, opened 2026-06-01
Feature: Extending init for artifact directory to support writing to another folder

CyrilCermak:
# This PR fixes one of the most critical problems for the CI pipelines. 

## Background:
ProcessInfo.processInfo.environment is not supported for multiple test targets by xctestplan, therefore, one can't propagate the EVN variable from the xctestplan to redirect the snapshots folder. 

## Usage:
By passing in the directory, the failed snapshots will be recorded into a different folder, therefore, repetition mode will no longer let the test succeed. As of now, when repetition mode is on, the first test fails and regenerates the snapshots, second try will load the regenerated snapshot and let the test pass. With recording the failed snapshots to a different folder will allow you to repeat the snapshots and they will be always reading the original snapshot. 

Furthermore, the folder with failed snapshots can then be used as an upload source of truth of the failing snapshots. Via scripting you can then replace the failing snapshots downloaded from the CI workflow locally to double check the changes and push it back.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -537,8 +537,11 @@ public func verifySnapshot<Value, Format>(
       }
 
       if record == .failed {
-        try recordSnapshot(writeToDisk: true)
-        failureMessage += " A new snapshot was automatically recorded."
+        let shouldWriteToDisk = artifactsDirectory == nil
+        try recordSnapshot(writeToDisk: shouldWriteToDisk)
+        if shouldWriteToDisk {
+          failureMessage += " A new snapshot was automatically recorded."
+        }
       }
 
       return """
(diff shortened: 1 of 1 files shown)


OlafAndreas:
I'm guessing this will provide the requested feature as described here https://github.com/pointfreeco/swift-snapshot-testing/discussions/960

Also noting that the blocker stated [here](https://github.com/pointfreeco/swift-snapshot-testing/discussions/960#discussioncomment-12540488) is implemented and merged https://github.com/swiftlang/swift-evolution/pull/2719

CyrilCermak:
> I'm guessing this will provide the requested feature as described here #960
> 
> Also noting that the blocker stated [here](https://github.com/pointfreeco/swift-snapshot-testing/discussions/960#discussioncomment-12540488) is implemented and merged [swiftlang/swift-evolution#2719](https://github.com/swiftlang/swift-evolution/pull/2719)

Yes exactly, this does not address the SwiftTesting, it is just a way how you can pass down the path where the snapshots will be stored. It's useful on the CI when you want to have your tests run up to N times, so that the failing snapshot does not re-generate itself in place but places the snapshot into the directory therefore the snapshot will be still failing N times.

PR #1101, opened 2026-06-02
Fix watchOS compile error: 'Image' not in scope

lapsefirsts8c:
**Problem**
  
  When compiling SnapshotTesting for watchOS, the build fails with:
  
  _error: cannot find 'Image' in scope_
  
  This occurs in `AssertSnapshot.swift` because:
  
  1. The `#available` check didn't include watchOS, so the compiler still evaluated the code path on watchOS but without the correct availability context.
  2. The Image typealias (mapping to UIImage) was never defined for the watchOS platform in `Internal.swift`.
  
  **Fix**
  
  - Added watchOS 7.0 to the #available check in AssertSnapshot.swift so the availability gate is correct on watchOS.
  - Added a `#elseif os(watchOS)` block in Internal.swift that imports UIKit and defines typealias Image = UIImage.
  
**Testing**
  
  Verified the project compiles successfully for watchOS targets after the change.

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -641,7 +641,7 @@ enum File {
   ) {
     #if !os(Android) && !os(Linux) && !os(Windows)
       #if compiler(>=6.3) && (canImport(UIKit) || canImport(AppKit))
-        if #available(iOS 14.0, tvOS 14.0, macOS 11.0, *),
+        if #available(iOS 14.0, tvOS 14.0, macOS 11.0, watchOS 7.0, *),
           name.hasSuffix(".png"),
           let image = Image(data: data)
         {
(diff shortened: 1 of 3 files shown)


Brett-Best:
I pulled this down and `swift-snapshot-testing/Tests/SnapshotTestingTests/SwiftTestingTests.swift` doesn't compile for watchOS.

lapsefirsts8c:
> I pulled this down and `swift-snapshot-testing/Tests/SnapshotTestingTests/SwiftTestingTests.swift` doesn't compile for watchOS.

Opps, pushed a fix, added `&& !os(watchOS)` to exclude view/image testing

msrutek-paylocity:
Hi @mbrandonw , would it be possible to take a look at this? This is preventing us from updating to a newer version. Thanks!

mbrandonw (approved):
Thanks @lapsefirsts8c !

PR #1106, opened 2026-07-02
Preserve the content type of failure diff attachments

### LINKED ISSUE #1107: Failure diff attachments lose their image content type in XCTest results
## Description

Since the `DiffAttachment` refactor, when a snapshot assertion fails under XCTest, the diff artifacts are attached to the test result as raw data with no content type:

```swift
case .data(let data, let name):
  let attachment = XCTAttachment(data: data)  // no uniformTypeIdentifier
  attachment.name = name
  activity.add(attachment)
```

[AssertSnapshot.swift#L510-L513 @ 1.19.2](https://github.com/pointfreeco/swift-snapshot-testing/blob/ad5e3190cc63dc288f28546f9c6827efc1e9d495/Sources/SnapshotTesting/AssertSnapshot.swift#L510-L513)

`XCTAttachment(data:)` with no UTI defaults the payload to `application/octet-stream`. Tools that consume `.xcresult` bundles (e.g. [XCTestHTMLReport](https://github.com/XCTestHTMLReport/XCTestHTMLReport)) only render an `` for attachments whose type is `image/*`, so the `reference`/`failure`/`difference` images are listed in the report but no picture is shown. `xcrun xcresulttool export attachments` likewise exports them as untyped data instead of `.png`.

Before the refactor the image strategies attached diffs via `XCTAttachment(image:)`, which carried the proper image type, so this is a regression for anyone publishing HTML reports from CI.

## To Reproduce

1. Make any image snapshot test fail with swift-snapshot-testing 1.19.x under XCTest.
2. Export attachments: `xcrun xcresulttool export attachments --path Test.xcresult --output-path out`.
3. The reference/failure/difference attachments are exported without an image type; XCTestHTMLReport shows empty attachment slots.

## Expected behavior

Diff attachments carry their image content type (e.g. `public.png`) so `.xcresult` consumers can render them.

## Environment

- swift-snapshot-testing 1.19.2
- Xcode 26.3, iOS 26.1 simulator

alexey1312:
Fixes #1107

## Summary

Since the `DiffAttachment` refactor, failure diff artifacts are attached to XCTest results via `XCTAttachment(data:)`, which defaults the payload type to `application/octet-stream`. Tools that consume `.xcresult` bundles (e.g. [XCTestHTMLReport](https://github.com/XCTestHTMLReport/XCTestHTMLReport)) only render attachments whose type is `image/*`, so the reference/failure/difference images no longer show up in HTML reports — the attachments are listed, but no picture is displayed.

## Fix

Derive the uniform type identifier from the artifact's file name (`reference.png`, `failure.png`, `difference.png`, `difference.patch`, …) and pass it to `XCTAttachment(data:uniformTypeIdentifier:)`, falling back to the previous behavior when the type cannot be resolved.

- `UniformTypeIdentifiers` is imported behind `#if canImport(...)` and used behind an `#available` check, matching the package's minimum deployment targets (iOS 13 / macOS 10.15 / tvOS 13 / watchOS 6).
- The swift-testing path is unaffected: `recordSwiftTestingAttachment` already preserves the file-name extension.

## Verification

Exported attachments from an `.xcresult` produced with this change: `xcrun xcresulttool export attachments` now exports the image diffs as `.png` (previously exported as untyped data).

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -7,6 +7,10 @@ import XCTest
   import AppKit
 #endif
 
+#if canImport(UniformTypeIdentifiers)
+  import UniformTypeIdentifiers
+#endif
+
 #if canImport(Testing)
   import Testing
 #endif
@@ -508,10 +512,14 @@ public func verifySnapshot<Value, Format>(
                   case .xcTest(let attachment):
                     activity.add(attachment)
                   case .data(let data, let name):
-                    let attachment = XCTAttachment(data: data)
+                    let attachment: XCTAttachment
+                    if let identifier = uniformTypeIdentifier(forFileName: name) {
+                      attachment = XCTAttachment(data: data, uniformTypeIdentifier: identifier)
+                    } else {
+                      attachment = XCTAttachment(data: data)
+                    }
                     attachment.name = name
                     activity.add(attachment)
-                    break
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


alexey1312:
Note on the failing **Android (6.2)** job: the 30 test failures are all `No reference was found on disk. Automatically recorded snapshot: file:///data/local/tmp/android-xctest/__Snapshots__/…` — the reference snapshots from the repo are not available on the emulator, so the assertions fall back to record mode and fail. None of the code in this PR executes on Android: the changed attachment block is inside the existing `#if !os(Linux) && !os(Android) && !os(Windows)` guard, and the new helper is only reachable from it (`UniformTypeIdentifiers` is additionally behind `#if canImport`). Linux and macOS jobs are green.

PR #1109, opened 2026-07-17
Prevent image diff crashes across pixel formats

### LINKED ISSUE #1097: normalizedComponentDiff can crash with SIGSEGV when generating the diff image after a failed comparison (1.19.2)
**Describe the bug**

When an image-based snapshot comparison fails, the library attempts to generate a visual diff via `normalizedComponentDiff(_:_:)` (introduced in #1028, first shipped in
  1.19.2). That function can crash with `EXC_BAD_ACCESS (SIGSEGV)` / `KERN_INVALID_ADDRESS` while reading pixel bytes, turning a normal `XCTFail` into a process crash that
  obscures the underlying test failure.
  
  The bug appears to be in how `normalizedComponentDiff` indexes the pixel buffers obtained from `cgImage.dataProvider?.data`.

  ## Affected version
  
  - `1.19.2` (current latest release)
  - `main` at time of writing — `e819244` is formatting-only, and no fix has landed since #1028.

  ## Root cause

  In `Sources/SnapshotTesting/Snapshotting/UIImage.swift`, `normalizedComponentDiff(_:_:)` (around lines 197–251) iterates pixel bytes like this:

  ```swift
  let oldBytes = CFDataGetBytePtr(oldData)!
  let newBytes = CFDataGetBytePtr(newData)!
  ...
  while index < pixelCount {
      defer { index += 1 }
      let pixelOffset = index * imageContextBytesPerPixel   // = index * 4

      let rOld = Int16(oldBytes[pixelOffset])
      ...
      let rNew = Int16(newBytes[pixelOffset])               // crashes here
      let gNew = Int16(newBytes[pixelOffset + 1])
      let bNew = Int16(newBytes[pixelOffset + 2])
      let aNew = Int16(newBytes[pixelOffset + 3])
      ...
  }
  ```

  The function assumes:
  
  1. **4 bytes per pixel** for both `oldData` and `newData` (`pixelOffset = index * imageContextBytesPerPixel`, then reads `+0..+3`).
  2. **No row padding** — i.e. `bytesPerRow == width * 4`.

  Neither assumption is guaranteed for `CFData` returned by `CGImage.dataProvider`:

  - `newCgImage` here is produced via `UIImage(data: new.pngData()!)?.cgImage` — a PNG round-trip. The decoded image may come back as RGB (3 bytes/pixel) rather than RGBA when
  the source is fully opaque, or with a different `bytesPerRow` than `width * bitsPerPixel/8` (CGImage is free to align rows).
  - `oldCgImage` is loaded directly from disk and is not normalized to the same format.

  The existing guard only checks `width` and `height` equality — not `bytesPerRow`, `bitsPerPixel`, or `CFDataGetLength`.

  When the actual buffer is smaller than `pixelCount * 4` (e.g. RGB at 3 bytes/pixel: `pixelCount * 3` bytes, ~25% smaller), the loop reads past the end of the malloc region
  near the last pixels and the process crashes.

  ## Observed crash shape (from a real crash report)

  - `EXC_BAD_ACCESS (SIGSEGV)` / `KERN_INVALID_ADDRESS`
  - Faulting address is exactly one byte past the end of an 8 MB `MALLOC_LARGE` region.
  - Reference image dimensions in this case were `1206 × 2622` RGBA → expected `~12 MB` if 4 bpp, but the actual buffer was `~8 MB`, consistent with a 3-bpp re-decode.
  - Top of stack:
    - `normalizedComponentDiff(_:_:)` (UIImage.swift, the `Int16(newBytes[pixelOffset])` line)
    - `diff(_:_:)`
    - The `image(precision:perceptualPrecision:scale:)` `Diffing` closure
    - `verifySnapshot(...)`
  - The comparison had already determined the snapshot didn't match; the crash happens **only** on the failure-reporting path while building the diff image.

  ## Minimal repro idea

  Any test that produces a failed image comparison where `new.pngData()` round-trips to a CGImage with a different pixel format than `oldCgImage` should reproduce. A synthetic
  repro:

  1. Save a reference PNG that decodes as RGBA (e.g. with at least one non-opaque pixel).
  2. In the test, render a slightly different image that is fully opaque (so PNG re-encode + decode lands at RGB / 3 bpp).
  3. Run `assertSnapshot(of: image, as: .image)` against that reference.

  Comparison fails → `diff()` → `normalizedComponentDiff()` → crash.

  ## Severity
  
  This converts a normal test failure into a process crash:

  - The user never sees the `XCTFail` message or attached diff.
  - On CI, the test runner reports a crash rather than a snapshot mismatch, masking the actual visual regression.
  - It is also non-obvious to diagnose: the stack trace points at the diff generator, not at the test that "really" failed.

  ## Suggested fix

  A few possible safeguards in `normalizedComponentDiff`, any of which would prevent the crash:

  1. **Re-decode through a known format**, the same way `compare(_:_:precision:perceptualPrecision:)` does — i.e. draw both `oldCgImage` and `newCgImage` into a `CGContext`
  with `imageContextBytesPerPixel`/`imageContextBitsPerComponent`/`premultipliedLast` and operate on those normalized buffers. This is the most robust option and matches the
  rest of the file.
  2. **Respect `cgImage.bytesPerRow`** when indexing instead of assuming `width * 4`: walk row-by-row using `rowStart = y * bytesPerRow`, `pixelOffset = rowStart + x *
  bytesPerPixel`.
  3. **Bail out (return `nil`) if the buffer is unexpectedly small**, e.g. by checking `CFDataGetLength(oldData) >= height * bytesPerRow` and same for `newData`. Falling back
  to `blendModeDiff` is preferable to crashing.

  Option 1 also makes `normalizedComponentDiff` independent of whatever format the original PNGs happened to be stored in, which seems closer to its intent.

**To Reproduce**
Zip up a project that reproduces the behavior and attach it by dragging it here.

```swift
// And/or enter code that reproduces the behavior here.

```

**Expected behavior**
Give a clear and concise description of what you expected to happen.

**Screenshots**

**Environment**
  - swift-snapshot-testing version [e.g. 1.9.0]
  - Xcode [e.g. 13.2]
  - Swift [e.g. 5.5]
  - OS: [e.g. iOS 15]

**Additional context**
Add any more context about the problem here.

thliu21:
## What

Normalize both images into the same 8-bit RGBA bitmap format before calculating the component diff, and add a regression test that compares images backed by 24-bit RGB and 32-bit RGBA data.

## Why

Generating a diff image for a failed snapshot comparison should report the mismatch and its attachments. It should not crash the test process when the reference and failure images use different underlying pixel formats.

## Root cause

`normalizedComponentDiff` read bytes directly from each `CGImage` data provider while indexing both buffers as tightly packed four-byte RGBA pixels. Core Graphics does not guarantee that provider data uses that pixel format or row layout, so a three-byte RGB buffer could be read past its end.

The updated implementation draws both images through the existing known-format bitmap context and performs the diff on those normalized buffers.

## Impact

Snapshot mismatches whose images decode to different pixel formats now produce the normal reference, failure, and difference attachments instead of risking `EXC_BAD_ACCESS` during failure reporting.

## Checks

- Xcode 26.5, iOS 26.5 simulator:
  `xcodebuild test -scheme swift-snapshot-testing-Package -destination 'platform=iOS Simulator,id=3BC2ABB5-651E-429E-8983-8CDBEC035388' -only-testing:SnapshotTestingTests/SnapshotTestingTests/testImageDiffWithDifferentPixelFormats`
- Result: 1 test passed, 0 failures.
- `git diff --check`

Fixes #1097

### DIFF
--- a/Sources/SnapshotTesting/Snapshotting/UIImage.swift
+++ b/Sources/SnapshotTesting/Snapshotting/UIImage.swift
@@ -179,7 +179,7 @@
 
   private func diff(_ old: UIImage, _ new: UIImage) -> UIImage {
     normalizedComponentDiff(old, new)
-    ?? blendModeDiff(old, new)
+      ?? blendModeDiff(old, new)
   }
 
   private func blendModeDiff(_ old: UIImage, _ new: UIImage) -> UIImage {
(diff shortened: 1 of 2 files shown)


alessdiimperio:
@mbrandonw  any chance to get this fix reviewed?
downgraded back after updating to 1.19.x due to this crash. its working on 1.18.7 but upgrading to 1.19.4 crashes when we diff

mbrandonw:
Hi @thliu21 and @alessdiimperio, unfortunately it's clear this whole process is just a game of whack-a-mole. We get one AI pull request saying something is fixed only for another AI pull request to say it broke something and then _this_ is the real fix.

If @CraigSiemens wants to take a look at this to vet (since he opened #1028) we will be happy to consider, but we are in no rush to just merge every PR that claims to fix something. Over and over again it only breaks something else. 

One thing we've mentioned repeatedly is that there is no impediment to maintaining your own snapshot strategies. There is no need to fork the library. You can create a snapshot strategy just as we have in the library that fixes your problems, and maintain that separate from the library. Have you tried that?

CraigSiemens:
@mbrandonw I can try looking into it some more. I haven't run into the issue while using the repo myself and just applying some simple changes to introduce a failing test haven't resulted in a crash for me.

Unfortunately the original issue doesn't contain repro steps, only an idea for possible repro steps.

> Any test that produces a failed image comparison where `new.pngData()` round-trips to a CGImage with a different pixel format than `oldCgImage` should reproduce.

The source of the "different pixel format" is the important missing detail. If we could get more details about how to reproduce the issue, that would help a lot.

Some possibilities I can think of are:
- a snapshot changing to/from being fully opaque
- comparing a UIKit snapshot to a SwiftUI snapshot
- comparing an opaque UIImage with a transparent one
- some external tool (like imageoptim) removing the alpha component from the saved reference images

alessdiimperio:
i'll be honest to say i haven't tried this fix, i just assumed that this was a broad issue that was easily reproducible.

Would it perhaps help if i run a crashing test on the latest and try to extract and provide the reference and failing image so you can diff and test and see if you can trigger the crash with the provided images?

mbrandonw:
A repro is of course always a helpful thing to provide. Without that we can't really make any changes to the library.

alessdiimperio:
not sure if i'll be able to have a reproducible test target but might be able to give you two images that you can try and diff on locally and see if you get a crash 🤔 

i'll try to get it done ASAP

alessdiimperio:
[images.zip](https://github.com/user-attachments/files/31339785/images.zip)
zipped up a dump of the old vs new image... that is causing the crash... 

seems as though the "new" is completely blank, not sure if it's empty or missing data but maybe you know more about png data and can deduce what's causing the crash if you can replicate the crash using these two images

old:    
1206x2622 bpp:32 bpc:8 rowBytes:4824 alpha:1 order:0

new:
1206x2622 bpp:16 bpc:8 rowBytes:2432 alpha:1 order:0

thliu21:
Thanks — the supplied files make this reproducible. The old image decodes as 32-bpp RGBA (12,648,528 bytes), while the new image is 16-bpp gray+alpha (6,324,264 bytes). On `1bc16f4`, a focused test using those exact files crashes with SIGSEGV; on `d9691465`, it passes. I also updated the deterministic regression to exercise 32-bpp versus 16-bpp without adding large binary fixtures. The focused iOS Simulator test passes 1/1.

CraigSiemens:
I found the same thing as well. It appears the iOS image renderer automatically chooses a greyscale pixel format when the image only contains greyscale values.

I was able to reproduce it with the following test. To make it reproduce the crash 100% of the time, it needs the address sanitizer enabled.

```swift
  func testImageDiffWithDifferentColorSpaces() throws {
    #if os(iOS) || os(tvOS)
      let size = CGSize(width: 100, height: 100)
      let renderer = UIGraphicsImageRenderer(size: size)
      
      func snapshot(of color: UIColor) -> UIImage {
        renderer.image { context in
          color.setFill()
          context.fill(.init(origin: .zero, size: size))
        }
      }
      
      let greyscaleSnapshot = snapshot(of: .white)
      let rgbSnapshot = snapshot(of: .red)
      
      let difference = Diffing<UIImage>.image.diffV2(greyscaleSnapshot, rgbSnapshot)

      XCTAssertEqual(difference?.0, "Newly-taken snapshot does not match reference.")
      XCTAssertEqual(difference?.1.count, 3)
    #endif
  }
```

I believe this PR does solve the issue. The only downside I'm seeing is that the images are rendered to the context multiple times when they don't match. Once in `compare` and again in `normalizedComponentDiff`. 

I started working on a refactor to remove that repetition but ran out of time this weekend. This is what I got so far 

https://github.com/pointfreeco/swift-snapshot-testing/compare/main...CraigSiemens:swift-snapshot-testing:image-diff-refactor

thliu21:
Thanks for reproducing this independently and for confirming that the PR fixes the crash. I traced the mismatch path and agree about the duplication: `compare` renders the reference, the new image, and the PNG-round-tripped new image; `normalizedComponentDiff` then PNG-round-trips the new image again and re-renders both normalized inputs. For the 1206x2622 repro, each RGBA buffer is 12,648,528 bytes, so that second pair alone writes another ~24.1 MiB, besides the extra PNG encode/decode. The focused regression also passes under Address Sanitizer (1/1).

Your draft removes that work, but it also changes dimension diagnostics and render-failure fallback behavior. Would you prefer keeping #1109 scoped to the crash and deterministic regression, or should I prepare a separate behavior-preserving buffer-reuse change for review?

PR #1112, opened 2026-07-25
refactor: skip the deprecated middle man

MojtabaHs:
The deprecated `record` property had an auto-fixing deprecation message that pointed users to **another** deprecated property, which was not the intended migration path. (IMO)

This change updates the deprecation message to guide users directly toward the recommended `withSnapshotTesting` API.

The implementation remains untouched to preserve a single source of truth and avoid duplicating the fallback logic.


mbrandonw (approved):
Thanks @MojtabaHs!

PR #1113, opened 2026-07-31
Fix OSS toolchain builds.

### LINKED ISSUE #1085: Build failure with Swift 6.3 open-source toolchain
**Describe the bug**

swift-snapshot-testing fails to build with the open-source `swift-6.3-RELEASE` toolchain. Builds fine with Xcode's Swift 6.2.4. I haven't tested Xcode 26.4 (requires macOS 26).

```
AssertSnapshot.swift:648:11: error: generic struct 'Attachment' requires that '_AttachableImageWrapper<NSImage>' conform to 'Attachable'
AssertSnapshot.swift:648:22: error: static method 'record(_:named:as:sourceLocation:)' requires that '_AttachableImageWrapper<NSImage>' conform to 'AttachableWrapper'
AssertSnapshot.swift:648:22: error: static method 'record(_:named:as:sourceLocation:)' requires that 'NSImage' conform to 'AttachableAsImage'
AssertSnapshot.swift:652:7: error: generic struct 'Attachment' requires that 'Data' conform to 'Attachable'
```

**To Reproduce**

[Install Swiftly](https://www.swift.org/install/) (swift.org's toolchain manager), then:

```
swiftly install 6.3.0
swiftly use 6.3.0
swift build --build-tests
```

(`swift build` alone succeeds — `canImport(Testing)` is false without `--build-tests`.)

**Environment**

- swift-snapshot-testing: main (ad5e319)
- Xcode 26.3.0 (Swift 6.2.4 — builds fine)
- Swift: `swift-6.3-RELEASE` toolchain from swift.org
- OS: macOS 15.7.5 (arm64)

**Additional context**

`Data: Attachable` and `NSImage: AttachableAsImage` live in cross-import overlay modules (`_Testing_Foundation`, `_Testing_AppKit`). SwiftPM only passes `-enable-cross-import-overlays` to test targets ([source](https://github.com/swiftlang/swift-package-manager/blob/dda772607e25f062942f4537b125a2a4edd4f5dc/Sources/Build/BuildDescription/SwiftModuleBuildDescription.swift#L1150)), so the `SnapshotTesting` library target can't see them.

The failure only manifests on Apple platforms because the `Attachment.record` calls are inside `#if !os(Android) && !os(Linux) && !os(Windows)` guards — though those guards may be unnecessary for the Swift Testing path, since `Attachment` is cross-platform unlike the `XCTContext`/`XCTAttachment` APIs they were originally protecting.

mbrandonw:
Fixes #1085

Currently SnapshotTesting doesn't build with OSS toolchains because cross-import overlays for Foundation, UIKit and AppKit are not automatically provided, and so the various `Attachable` conformances are not found. We can manually import these overlays as a very easy fix, but the overlays are only available on newer deployment targets. Up for discussion of whether or not we want to bump deployment targets (we still support over 3 years of platforms).

### DIFF
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -11,9 +11,6 @@ import XCTest
   import Testing
 #endif
 
-#if canImport(_Testing_Foundation)
-  import _Testing_Foundation
-#endif
 #if canImport(UIKit) && canImport(_Testing_UIKit)
   import _Testing_UIKit
 #elseif canImport(AppKit) && canImport(_Testing_AppKit)
@@ -643,15 +640,12 @@ enum File {
   ) {
     #if !os(Android) && !os(Linux) && !os(Windows)
       #if compiler(>=6.3) && (canImport(UIKit) || canImport(AppKit))
-        if #available(iOS 14.0, tvOS 14.0, macOS 11.0, *),
-          name.hasSuffix(".png"),
-          let image = Image(data: data)
-        {
+        if name.hasSuffix(".png"), let image = Image(data: data) {
           Attachment.record(image, named: name, as: .png, sourceLocation: sourceLocation)
           return
         }
       #endif
-      Attachment.record(data, named: name, sourceLocation: sourceLocation)
+      Attachment.record([UInt8](data), named: name, sourceLocation: sourceLocation)
     #endif
   }
 #endif
--- a/Sources/SnapshotTesting/AssertSnapshot.swift
+++ b/Sources/SnapshotTesting/AssertSnapshot.swift
@@ -11,6 +11,15 @@ import XCTest
   import Testing
 #endif
 
+#if canImport(_Testing_Foundation)
+  import _Testing_Foundation
+#endif
+#if canImport(UIKit) && canImport(_Testing_UIKit)
+  import _Testing_UIKit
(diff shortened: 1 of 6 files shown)


Sources/SnapshotTesting/Snapshotting/NSBezierPath.swift
```diff
@@ -59,14 +59,14 @@
       let namesByType: [NSBezierPath.ElementType: String] = [
         .moveTo: "MoveTo",
         .lineTo: "LineTo",
-        .curveTo: "CurveTo",
+        .cubicCurveTo: "CurveTo",
```
mbrandonw: Deprecation warning fixes...

Sources/SnapshotTesting/AssertSnapshot.swift
```diff
@@ -574,17 +583,10 @@ func sanitizePathComponent(_ string: String) -> String {
 }
 
 #if !os(Android) && !os(Linux) && !os(Windows)
-  import CoreServices
+  import UniformTypeIdentifiers
```
mbrandonw, line 586: Deprecation warning fixes...

mbrandonw:
We are going to close this since it unfortunately will break #1067, and it's hard to support it + fix Swiftly at the same time. However, once we can drop the iOS 15 generation of platforms (soon) we will be able to bring this fix back.

PR #1120, opened 2026-09-16
Fix perceptual image comparison crash on macOS 27

adamalix:
On macOS 27, perceptual comparison crashes when two images differ instead of returning a snapshot mismatch. `CIAreaAverage` and `CIAreaMaximum` no longer work with `CGRect` representation used for `inputExtent`. I fixed it by passing a `CIVector(cgRect:)`, which both filters accept. 

This is the list of simulators & hardware I verified this on:

- iOS 27.0 simulator
- iOS 26.5 simulator
- macOS 27 MacBook Pro M5 Pro
- macOS 26.6 MacBook Pro M4 Pro
- tvOS 27.0 simulator

**NOTE:** I drove this change with a Codex agent, but have verified it myself and understand it.


temoki:
Another data point for this, from a different project that hit the same crash: **it is not specific to macOS 27 — the variable is the toolchain, not the OS.** Holding the OS fixed at macOS 26.6.2 (25G83) and moving only Xcode 26.6 → 27.0 (27A266a) turns the same `CIAreaAverage` call from passing into throwing.

| OS | Toolchain | `CIAreaAverage` + `CGRect` | `CIAreaAverage` + `CIVector` |
|---|---|---|---|
| macOS 26.6.2 (25G83) | Xcode 26.6 | passes | passes |
| macOS 26.6.2 (25G83) | Xcode 27.0 (27A266a) | **throws** | passes |
| macOS 27.0 (26A5406e) | Xcode 27.0 (27A266a) | **throws** | passes |

Minimal reproduction, with the table above produced by its CI: https://github.com/temoki/snapshot-testing-macos-extent-crash — one test target isolates the `inputExtent` parameter with no dependency on this library, the other goes through `Diffing<NSImage>.image(precision:perceptualPrecision:).diff` on two images that differ by a single pixel.

Two details that may be worth adding to the PR description:

- On macOS 26 built with Xcode 27, the library's own call path does not reach the evaluation that throws, so only a direct `CIAreaAverage` call shows it. On macOS 27 the library path reaches it too. That is why this looks macOS-27-specific from inside a test suite, even though the wrong parameter type is rejected on both.
- **Under swift-testing the whole test process dies** (signal 6, every other test in the bundle lost), because nothing catches the Objective-C exception. Under XCTest it is caught and reported as a single test failure. Our suites are swift-testing, so one differing snapshot took out the entire bundle.

Your fix is exactly what we arrived at independently (`CIVector(cgRect:)` in both `applyingAreaAverage()` and `applyingAreaMaximum()`), and the `CIVector` column passes in every combination we measured. +1 from us.

rcarver:
Confirmed fixes running tests here, thanks!

rcarver:
Followup here: running the `swift-snapshot-testing` tests in Xcode 27 doesn't show this failure at `testImagePrecision` because the existing snapshots match exactly, so they don't actually exercise the precision-checking paths.

To verify the fix, I modified the fixture image in place forcing a precision check, which fails with error:

```
.../swift-snapshot-testing/Sources/SnapshotTesting/Snapshotting/UIImage.swift:388 testImagePrecision(): -[NSConcreteValue CGRectValue]: unrecognized selector sent to instance 0x7562e38640 (NSInvalidArgumentException)
```

After that, applying the patch causes the snapshots to fail normally, re-record, and then pass.

adamalix:
thanks @mbrandonw, really appreciate it.

temoki:
Confirmed resolved in 1.19.6 — thanks for the fix.

